PR #94, opened 2019-10-04
Don't add empty fields to the letter

ggangix:
Checking if the fields are empty and don't add the title/value to the final letter. If all the values are empty we have set to true the flag all_fields_empty.

To be done: Open an issue (?) to use the flag in the template.

### DIFF
--- a/src/Utility/RequestLetter.js
+++ b/src/Utility/RequestLetter.js
@@ -140,6 +140,7 @@ export default class RequestLetter {
 
         flags = {
             erase_some: !request_object.erase_all,
+            all_fields_empty: !id_data.formatted,
             erase_all: request_object.erase_all,
             data_portability: request_object.data_portability,
             runs: request_object.recipient_runs ? request_object.recipient_runs.length > 0 : false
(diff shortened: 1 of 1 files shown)


baltpeter:
This looks great, thank you very much!

Paging @zner0L for the review as he is (mostly) responsible for this code.

baltpeter:
Deploy preview for *datenanfragen* failed.

Built with commit 4e184312a9ec9e6debdc2b88eb3469725bc8170c

https://app.netlify.com/sites/datenanfragen/deploys/5d978aeeecb52e00087d45c7

baltpeter:
I have created datenanfragen/data#238 for adding the flags to the templates. We definitely have to merge this PR first anyway, though.

As for Netlify complaining about the build failing, this doesn't seem to be caused by this PR. I have created #95 for that.

ggangix:
Yes you right, doesn't look like a error caused by this PR. Let's try to fix in the other issue.

src/Utility/RequestLetter.js
```diff
@@ -173,30 +174,33 @@ export default class RequestLetter {
         let name = '';
 
         request_data.forEach(item => {
```
zner0L: Now that I look at it again I feel like this could have been done a little more elegantly using `Array.prototype.reduce()`. Oh, well.

baltpeter: Yeah, I also have quite a few places where I would much rather use a `.map()` or `.reduce()` instead of an ugly `forEach`. We'll get there eventually. :D

baltpeter:
Apparently neither me nor @zner0L were paying proper attention: Our template engine can only check whether a flag is true (not false), so actually we need to invert the `all_fields_empty` flag.

Done in a2a1a85, it is now called `has_fields`.

PR #101, opened 2019-10-04
Prompt for removal confirmation only when input contains user entered…

sibis:
Current Scenario:  When removing the user-entered data, user is alerted regarding the action even when the input field does not have any data in it.

Fix: During the removal, check has been made whether the user has entered any data in that field and action will be taken based on that. If it contains it will throw alert to confirm else removed without alert.

### DIFF
--- a/src/Components/Generator/DynamicInputContainer.js
+++ b/src/Components/Generator/DynamicInputContainer.js
@@ -225,16 +225,18 @@ export default class DynamicInputContainer extends preact.Component {
                 return prev;
             });
             this.pushStateUp();
-        } 
+        }
     }
 
     // returns boolean whether the field has value based on its type
     static isFieldEmpty(field_value) {
-        if((typeof field_value == 'string') && (field_value)){
-           return false;  
-        } else if(typeof field_value == 'object') {
+        if (typeof field_value == 'string' && field_value) {
+            return false;
+        } else if (typeof field_value == 'object') {
             for (let [key, value] of Object.entries(field_value)) {
-                if((key != 'primary') && (value)) { return false; }
+                if (key != 'primary' && value) {
+                    return false;
+                }
             }
         }
         return true;
--- a/src/Components/Generator/DynamicInputContainer.js
+++ b/src/Components/Generator/DynamicInputContainer.js
@@ -216,13 +216,28 @@ export default class DynamicInputContainer extends preact.Component {
     }
 
     removeDynamicInput(event) {
-        if (window.confirm(t('confirm-input-remove', 'generator'))) {
+        // prompt only if respective field has value entered while removing
+        const field = event.target.getAttribute('rel');
+        const field_value = this.state.fields[field].value;
+        if (DynamicInputContainer.isFieldEmpty(field_value) || window.confirm(t('confirm-input-remove', 'generator'))) {
             this.setState(prev => {
                 delete prev.fields[event.target.getAttribute('rel')];
                 return prev;
             });
             this.pushStateUp();
+        } 
+    }
+
+    // returns boolean whether the field has value based on its type
+    static isFieldEmpty(field_value) {
+        if((typeof field_value == 'string') && (field_value)){
+           return false;  
+        } else if(typeof field_value == 'object') {
+            for (let [key, value] of Object.entries(field_value)) {
+                if((key != 'primary') && (value)) { return false; }
+            }
         }
+        return true;
     }
 
     getDataArray() {


baltpeter:
Deploy preview for *datenanfragen* failed.

Built with commit d97ba9f223319a048319bcf19d2120dbcd116b88

https://app.netlify.com/sites/datenanfragen/deploys/5d97b6142f8612000a331a6a

baltpeter:
Thank you, looks good to me! We'll just need to run Prettier over the file to make sure it adheres to our code style guidelines.

Paging @zner0L for review.

baltpeter:
Not sure if you've seen this already but feel free to signup to our Hacktoberfest event to get some stickers from us, if you want: https://www.datarequests.org/blog/hacktoberfest-2019/

sibis:
> Not sure if you've seen this already but feel free to signup to our Hacktoberfest event to get some stickers from us, if you want: https://www.datarequests.org/blog/hacktoberfest-2019/

Sure @baltpeter, Thanks :)

src/Components/Generator/DynamicInputContainer.js
```diff
@@ -216,13 +216,34 @@ export default class DynamicInputContainer extends preact.Component {
...
-            this.pushStateUp();
+        // prompt only if respective field has value entered while removing
+        if (!this.checkInputData(event)) {
+            this.confirmStateChange(event);
+        } else if(window.confirm(t('confirm-input-remove', 'generator'))) {
+            this.confirmStateChange(event);
+        }   
+    }
+
+    confirmStateChange(event) {
```
zner0L: I think this function name is misleading. Since this might delete data the naming should be very clear.

src/Components/Generator/DynamicInputContainer.js
```diff
@@ -216,13 +216,34 @@ export default class DynamicInputContainer extends preact.Component {
...
+            delete prev.fields[event.target.getAttribute('rel')];
+            return prev;
+        });
+        this.pushStateUp();
+    }
+
+    // returns whether the field has value based on its type
+    checkInputData(event) {
+        let val = event.target.getAttribute('rel');
+        let name = this.state.fields[val].value;
```
zner0L: These can be `const` but I guess the Prettier will take care of that. But the variable names are also misleading here. Just from reading the names I would assume `val` to contain the field value and `name` to contain its name. Maybe call them something like `field` and `field_value`. 

Also, to make this method more generic you could move the event to `removeDynamicInput()` and just pass down the field id.

src/Components/Generator/DynamicInputContainer.js
```diff
@@ -216,13 +216,34 @@ export default class DynamicInputContainer extends preact.Component {
...
-                delete prev.fields[event.target.getAttribute('rel')];
-                return prev;
-            });
-            this.pushStateUp();
+        // prompt only if respective field has value entered while removing
+        if (!this.checkInputData(event)) {
+            this.confirmStateChange(event);
+        } else if(window.confirm(t('confirm-input-remove', 'generator'))) {
+            this.confirmStateChange(event);
+        }
```
zner0L: I actually don't like to clutter this already cluttered class with another method of very limited use. Try grouping those together in one statement doing something like `!this.checkInputData(…) || window.confirm(…)`.

src/Components/Generator/DynamicInputContainer.js
```diff
@@ -216,13 +216,34 @@ export default class DynamicInputContainer extends preact.Component {
...
+    confirmStateChange(event) {
+        this.setState(prev => {
+            delete prev.fields[event.target.getAttribute('rel')];
+            return prev;
+        });
+        this.pushStateUp();
+    }
+
+    // returns whether the field has value based on its type
+    checkInputData(event) {
```
zner0L: Try hinting to the return value by naming the method in such a way that you can read it easily when it is used in other code, e.g. `isFieldEmpty()` (in this case the return values should obviously be switched).

src/Components/Generator/DynamicInputContainer.js
sibis: @zner0L Sure, I think what you have suggested totally makes sense. I will update this.

sibis:
@zner0L have updated the PR with all your suggestions. Please let me know if anything can be improved further. Thanks :)

src/Components/Generator/DynamicInputContainer.js
```diff
@@ -216,13 +216,28 @@ export default class DynamicInputContainer extends preact.Component {
...
             this.setState(prev => {
                 delete prev.fields[event.target.getAttribute('rel')];
                 return prev;
             });
             this.pushStateUp();
+        } 
+    }
+
+    // returns boolean whether the field has value based on its type
+    isFieldEmpty(field_value) {
```
zner0L: This can be static now.

sibis:
> Very cool! Just make `isFieldEmpty()` `static` and I will merge.

have taken care @zner0L. Thanks

sibis:
> Very cool! Just make `isFieldEmpty()` `static` and I will merge.

have taken care of, @zner0L. Thanks

PR #111, opened 2019-10-06
Created DropdownComponent.js and removed dropdown from baseof.html

thebhushanmhatre:
I haven't checked whether this works or not because I am not familiar with technological stack and there were many things I didn't understand.
So instead of further spending time on this, Its better if somebody else continues on this.
Apologies in advance if this doesn't work.

Thanks.


baltpeter:
Deploy preview for *datenanfragen* ready!

Built with commit 190f2f497c79159e714803e51d94c5a9c582006e

https://deploy-preview-111--datenanfragen.netlify.com

Issue #98, opened 2019-10-04
Allow deleting individual saved requests


baltpeter:
On the "My requests" page (`/my-requests`), there is currently only the option of deleting all saved requests.

There should be a delete button next to each request (in a new column), allowing the user to delete individual requests.  
We already have a delete button design (in the wizard, for example). This should be used here, as well.

![image](https://user-images.githubusercontent.com/4048809/66234802-263a6480-e6de-11e9-96dc-1c334be28d21.png)

In addition, the user should be able to delete multiple requests at once (the selected ones), just like they can export the selected requests.

MeisterRed:
Hi, I would like to work on this.

baltpeter:
Great, thank you!

MeisterRed:
Hey. I'm having some trouble with completing the issue due to it be a bit harder than expected. I'm not super experienced with react so I think its best someone else take the issue from here. My apologies

baltpeter:
No worries, thanks for trying!

xeluna:
Hello, I'd like to work on this. May I ask if there's a deletion message text that should be shown?

baltpeter:
@xeluna Thank you!

What do you mean by that? The specific text you use is not that important, we'll have to translate it anyway, so we can make changes then if necessary.

xeluna:
Okay, I can put a placeholder in the meantime.

PR #123, opened 2019-10-11
Add delete button for individual requests

xeluna:
#98 
Meine Anfragen · Datenanfragen de

For the object property name and the deletion confirmation messages, let me know if you would like them changed to something else. There probably is a better way to word the German version.

### DIFF
--- a/src/i18n/de.json
+++ b/src/i18n/de.json
@@ -154,6 +154,7 @@
         "cancel": "Abbrechen",
         "modal-clear-requests": "Damit werden alle gespeicherten Anfragen unwiderruflich von Deinem Computer gelöscht. Bist Du sicher?",
         "modal-clear-id_data": "Damit werden Deine gespeicherten Identifikationsdaten gelöscht. Bist Du sicher?",
+        "modal-delete-single-request": "Damit wird diese Anfrage unwiderruflich von Deinem Computer gelöscht. Bist Du sicher?",
         "clear-success": "Löschen abgeschlossen.",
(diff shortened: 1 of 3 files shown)


baltpeter:
Deploy preview for *datenanfragen* ready!

Built with commit 2927da6e17bced54af1c8446a1a923c820543653

https://deploy-preview-123--datenanfragen.netlify.com

xeluna:
Agh, I knew I forgot something. I've added the deletion from local storage.

zner0L:
@baltpeter @xeluna I know that the issue was asking something different but why not just add another button similar to the existing one on the bottom? I mean, you can already select individual requests and export them, we just need a delete action as well.

xeluna:
That's also another way to go about it. If that is what is preferred, I can change the implementation.

One reason I think for having the delete button inline is that the page has all the requests selected by default. This is a good default for exporting. But, for selection to delete, the user should have to select exactly what they want to avoid mistakenly deleting more than they intended.

For new users who are unfamiliar, there is the risk that they did not read the confirmation message correctly and end up deleting everything.

Of course this is from my own limited knowledge of how the website is used and I may be wrong.

baltpeter:
@zner0L My original idea was having both features. I thought I had also created an issue for deleting multiple requests but apparently not.

You are raising a very valid point though, @xeluna. I hadn't considered that.

Either way, I think we should definitely have the option of deleting individual requests. Unless you are opposed to that, @zner0L, I will merge this.

PR #132, opened 2019-10-15
Update README.md

IfeanyiEkperi:
Changed 'from Germany' to 'in Germany' (using from would mean the project is based in another country).

Changed 'who' to 'they' (Since GitHub CI and BrowserStack aren't humans but projects with contributors, using who is inappropriate, hence 'they' fits in well in the sentence.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 > This repository contains the source code for the Datenanfragen.de website.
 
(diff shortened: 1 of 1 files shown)


baltpeter:
Sorry but we don't agree that these changes add value.

IfeanyiEkperi:
Even if these changes wouldn't count on the Hacktoberfest challenge, I still feel they should be effected. Thank you

baltpeter:
As I said, we disagree. We don't differentiate between contributions that qualify for Hacktoberfest and ones that add value but don't. Anything that adds value to the project qualifies. We just don't think that is the case here.

As for the first change, we are **from** Germany. And the second one is nit-picking at best, without making the respective sentence any easier to understand.

PR #133, opened 2019-10-15
indicate email already sent

### LINKED ISSUE #92: Indicate when we consider a request done
When working with the wizard, it is sometimes easy to forget if you have sent the request that is currently open already (especially when pausing for a while).

Since we already have a point at which we consider a request 'done', we should simply indicate that to the user (by changing the color of the button or something similar).

peterjgrainger:
Fixes #92

# PDF

## Request not complete

image

## Request complete

image

# Email

## Request not complete

image

## Request complete

image


src/i18n/de.json
```diff
@@ -53,9 +53,11 @@
     "generator": {
         "reference": "Zeichen",
         "download-pdf": "PDF herunterladen",
+        "download-pdf-again": "Laden Sie das PDF erneut herunter",
```
zner0L: We use "Du" as pronoun in German.

zner0L: You could actually get rid of the pronoun, just write: "PDF erneut herunterladen"

res/icons/delivery-truck.svg
```diff
@@ -0,0 +1 @@
+<svg width="24" height="24" xmlns="http://www.w3.org/2000/svg" fill-rule="evenodd" clip-rule="evenodd"><path d="M5 18H4a1 1 0 01-1-1v-2h12V6h4.667c1.117 0 1.6.576 1.936 1.107.594.94 1.536 2.432 2.109 3.378.188.312.288.67.288 1.035V16c0 1.121-.728 2-2 2h-1a3.001 3.001 0 01-6 0h-4a3.001 3.001 0 01-6 0zm3-1.2a1.201 1.201 0 010 2.4 1.201 1.201 0 010-2.4zm10 0a1.201 1.201 0 010 2.4 1.201 1.201 0 010-2.4zM8 8H0V6h8V5H3V4a1 1 0 011-1h9a1 1 0 011 1v10H3v-3H1V9h7V8zm9 0v3h4.715l-1.427-2.496A.999.999 0 0019.42 8H17z"/></svg>
```
zner0L: This SVG has not defined a `viewBox`. Maybe that is the source of the error?

peterjgrainger: Unfortunately not. It happens when it is linted. See https://github.com/datenanfragen/website/issues/92#issuecomment-543612612

This will be an issue for all new svg added. I'll try and change the configuration for lint-staged but I haven't had anything to do with that before.

peterjgrainger: It's not just that svg I tried a few more and they don't work either. The paper plane works OK after the linting.

zner0L: OK, so I will try and fix the problem in `imagemin-lint-staged`, but until then I guess we should just disable `imagemin-lint-staged` for `.svg` altogether. So if you remove that in the `package.json`, `lint-staged` should leave the delivery truck be and the build should work. I will reimplement it as soon as I fixed the problem.

zner0L:
Cool. This seems to work. I fixed the problem and will push it to your branch and merge it then. Thank you!

Issue #93, opened 2019-10-03
Translate GDPR cheat sheet blog post from German to English


baltpeter:
We have a [GDPR cheat sheet blog post](https://www.datenanfragen.de/blog/dsgvo-glossar/) ([this file](https://github.com/datenanfragen/website/blob/master/content/blog/gdpr-cheat-sheet.de.md)) that is essentially just a short explanation of important GDPR terms and topics.

That blog post is currently only available in German. We should translate it to English.

drd-2rcl:
Hi, i would like to get this issue

baltpeter:
Sure, thank you!

baltpeter:
Hi! Just a quick note: We are now officially participating in the Hacktoberfest. Anyone who submits a PR to one of our repositories can get a free sticker set from us in addition to the regular swag by DigitalOcean.

Here's our blog post on the subject: https://www.datarequests.org/blog/hacktoberfest-2019/

MIP2000:
Hello, i translated the post! Check if everything is alright. Thanks a lot.

xeluna:
I think this issue can be closed since the PR has been merged. :)

baltpeter:
It certainly can. I am too used to that happening automatically from the commit messages. :D Unfortunately, we can't do that right now as we are running GitHub and a private GitLab instance in parallel with different issue numbering systems… Oh well.

PR #134, opened 2019-10-15
Translate GDPR Cheat Sheet

xeluna:
#93, since I think no one is working on it.

Let me know if there's other changes to the metadata required. There currently isn't a field specifically for the translator so I didn't add my name.

Any suggestions on how to word the translations better are welcome.

### DIFF
--- a/content/blog/gdpr-cheat-sheet.en.md
+++ b/content/blog/gdpr-cheat-sheet.en.md
@@ -47,7 +47,7 @@ What is important for you as a consumer: As a rule, the data processor is not th
 
 ### Data protection officer
 
-A data protection officer's task is to ensure that a company (or another organisation, such as a public authority) complies with data protection law. He is appointed by the company itself and must have a certain degree of independence in order to perform his duties.
+A data protection officer's task is to ensure that a company (or another organisation, such as a government agency) complies with data protection law. He is appointed by the company itself and must have a certain degree of independence in order to perform his duties.
 
 If a [controller](#data-controller) has appointed a data protection officer, this person should be your contact person for data protection-related enquiries, for example if you want to exercise your [rights](#rights).
 
@@ -110,7 +110,7 @@ You will find a comprehensive overview and a detailed explanation on this subjec
 
 ### Special categories of personal data
 
-Some data is more sensitive than others. While you may have no qualms about giving your name to a new app, it doesn't have to know your religious beliefs. The GDPR recognizes that these so-called "special categories of personal data" deserve and need special protection. Therefore, the [processing](#processing) of such data is only allowed under certain circumstances.
+Some data is more sensitive than others. While you may have no qualms about giving your name to a new app, it doesn't have to know your religious beliefs. The GDPR recognises that these so-called "special categories of personal data" deserve and need special protection. Therefore, the [processing](#processing) of such data is only allowed under certain circumstances.
 
 Concretely, this concerns the following data:
 
--- a/content/blog/gdpr-cheat-sheet.en.md
+++ b/content/blog/gdpr-cheat-sheet.en.md
@@ -1,13 +1,14 @@
 {
-    "title": "GDPR Cheat Sheet – An Explanation of the Important Terms and Concepts",
-    "heading": "GDPR Cheat Sheet<br><span class='color-gray-800' style='font-size: 0.75em;'>An Explanation of the Important Terms and Concepts</span>",
+    "title": "GDPR cheat sheet — an explanation of the important terms and concepts",
+    "heading": "GDPR cheat sheet<br><span class='color-gray-800' style='font-size: 0.75em;'>An explanation of the important terms and concepts</span>",
     "type": "blog",
     "slug": "gdpr-cheat-sheet",
     "date": "2019-02-16T16:44:39+02:00",
+    "last_edited": "2019-10-16T04:17:02.56+02:00",
     "description": "At first glance, the GDPR can appear confusing like any other legal topic. In this article, we briefly explain the most important terms and concepts related to the GDPR so that you can easily remain in control.",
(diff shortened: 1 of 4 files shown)


content/blog/gdpr-cheat-sheet.en.md
```diff
@@ -0,0 +1,139 @@
...
+* Work performance
+* Economic situation
+* Health
+* Personal preferences
+* Interests
+* Reliability
+* Behaviour
+* Location or movements
+
+A classic example is the mobile phone or credit card credit check that you need to pass before getting a contract. Many companies will forward your data to credit agencies such as [Schufa]({{< ref "schufa" >}}), [Creditreform Boniversum]({{< ref "creditreform-boniversum" >}}) or [infoscore]({{< ref "infoscore" >}}). These agencies will then determine how likely you are to fulfil your contractual obligations based on information such as your previous payment behaviour or the number of bank accounts that you have.
```
xeluna: Do you want these examples replaced with some non-German examples?

baltpeter:
Thank you, will take a closer look when I have the time. As for your two questions:

* I don't think we need to differentiate between the author and translator. Just add your name (alias is fine, of course) and an optional link to `config/_default/params.toml` and then reference that in the `authors` key of the post. :)
* Please do replace German examples with better alternatives for an English post (i.e. UK-specific or international ones, I suppose), yes.

content/blog/gdpr-cheat-sheet.en.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Data Minimisation
+
+Data minimisation is an important principle for data protection that the GDPR prescribes. The idea behind it is simple: Only the data necessary for carrying out the stated purpose should be collected and [processed](#processing). If a company wants to send you a newsletter, all it needs is your email address. Following the data minimisation principle, it should not collect any other data.
+
+This concept is important not least because of security reasons. We live in a time where users' private data are regularly leaked to the public because companies have not sufficiently secured them. Data that a company does not have cannot be lost.
+
+### Data Protection Officer
+
+A data protection officer's task is to ensure that a company (or another organisation, such as an authority) complies with data protection law. He is appointed by the company itself and must have a certain degree of independence in order to perform his duties.
```
baltpeter: The term `authority` may be a little misleading here. We usually use `public body` (although technically, that has a bit of a different meaning). `Government agency` also works fine.

content/blog/gdpr-cheat-sheet.en.md
```diff
@@ -0,0 +1,139 @@
...
+
+Our [Data Protection Supervisory Authority Finder]({{< ref "supervisory-authorities" >}}#finder) will help you find the supervisory authority responsible for you.
+
+### Personal Data
+
+The concept of personal data is the crux of the GDPR. Almost all of the regulations refer to it. As soon as information can be assigned to a specific person, it is considered personal data. A name is not necessarily required for the assignment. For example, an IP address, a user name, a cookie, or an identification number would suffice.
+
+The information that the IP address 31.41.59.26 called our [generator]({{< ref "generator" >}}) on 25 May 2018 would therefore be considered personal data. On the other hand, the information that on the same day in a bakery a third of the customers paid with cash is not.
```
baltpeter: Might just be me but I would associate calling with using a phone. :D How about `visited`?

content/blog/gdpr-cheat-sheet.en.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Rights
+
+The GDPR grants you extensive rights with regards to your [personal data](#personal-data). This starts with the right to information: If a [controller](#data-controller) wants to [process](#processing) your data, they must inform you about it and tell you how you can contact them if you have any questions.
+
+Even after the processing of your data has started, you still have numerous other rights. For example, you can ask what personal data is being processed and request a copy of it. You can also request for incorrect data to be corrected and under certain circumstances, even have your data deleted.
+
+You will find a comprehensive overview and a detailed explanation on this subject in our [article on your GDPR rights]({{< ref "your-gdpr-rights" >}}).
+
+### Data Controller
```
baltpeter: Not the first instance where this is relevant but here's where I noticed it: In the German post, I sorted the heading alphabetically. We don't need to do that here, though. I'll leave the decision to you.

content/blog/gdpr-cheat-sheet.en.md
xeluna: I went to look at how it was phrased and [here](https://gdpr-info.eu/art-37-gdpr/), it uses `public authority`. I did more research and it seems that the GDPR itself does not specify what this entails, and it is up to member states to define it exactly.

For the UK, it's broader than government agencies only, and using that might narrow the definition too much. Would it be too technical to leave it as `public authority`?

content/blog/gdpr-cheat-sheet.en.md
xeluna: `visited` would definitely work better, I was thinking too technically. :P

xeluna:
Here's a summary of the changes:
- Changed headings to sentence case
- Sorted terms alphabetically
- Replaced the credit agencies example with (UK's) Experian and Equifax
- Other suggestions from @baltpeter

content/blog/gdpr-cheat-sheet.en.md
baltpeter: `public authority` is definitely fine, too. I just want to avoid confusion with the supervisory authorities (although, to be fair, they will have their own DPOs, too).

xeluna: I decided to use `government agency` in the end, since, after all, it's meant to be one example. This would definitely make a clearer distinction from supervisory authorities.

baltpeter:
Awesome, let's get this merged.

Issue #91, opened 2019-10-03
Automatically crop the signature


baltpeter:
It would be great if we could crop the signature.

Whenever I tested the signature input, I found it quite hard to judge how much space was needed for my signature and often ended up with quite large margins (especially at the top and bottom) which look fairly ugly in the PDF.

Snaked06:
We can use Imagemagick To succesfully do this for example https://stackoverflow.com/questions/50187924/auto-crop-textsignature-from-image-and-change-background-color-using-imagemagi

baltpeter:
Thanks for the suggestion, I don't think ImageMagick works on the client-side, though. Besides, I would like to try and avoid pulling in that big of a dependency, the generator is already quite heavy.

This should be possible in plain JS using the canvas APIs.

merelj:
that sounds like a cool feature, I'd like to try :)

baltpeter:
Sure, thank you!

merelj:
@baltpeter 
opened [this pr](https://github.com/datenanfragen/website/pull/135), but still need some help with it :(

PR #135, opened 2019-10-15
Automatically crop the signature

merelj:
Should solve [this](https://github.com/datenanfragen/website/issues/91) issue... eventually.

For now I've figured out how to crop the canvas (thanks to stack overflow) and also figured out that it conflicts a bit with detectBlockedCanvasImageExtraction (which does not restore the original pixel color (at least for me in chrome)).

I store the result in `croppedValue` property on signature object since original `value` property is still used to store canvas state.

What I still have troubles with is getting the croppedValue to render in pdf instead of value; would be grateful for any pointers.

### DIFF
--- a/src/Components/Generator/SignatureInput.js
+++ b/src/Components/Generator/SignatureInput.js
@@ -72,7 +72,15 @@ export default class SignatureInput extends preact.Component {
                 const top = (this.canvas.height - img.height) / 2;
                 this.clear();
                 this.context.drawImage(img, left, top);
-                this.setState({ isEmpty: false });
+                this.setState({
+                    isEmpty: false,
+                    cropArea: {
+                        top,
+                        bottom: img.height + top,
+                        left,
+                        right: img.width + left
+                    }
+                });
                 this.handleChange();
             };
             img.src = signature.value;
@@ -101,8 +109,7 @@ export default class SignatureInput extends preact.Component {
     clear() {
         if (this.state.isEmpty) return;
         this.context.clearRect(0, 0, this.state.width, this.state.height);
-        this.setState({ isEmpty: true });
-        this.setState({ cropArea: this.initialCropArea });
+        this.setState({ isEmpty: true, cropArea: this.initialCropArea });
         this.handleChange();
     }
 
--- a/src/Components/Generator/SignatureInput.js
+++ b/src/Components/Generator/SignatureInput.js
@@ -9,6 +9,12 @@ export default class SignatureInput extends preact.Component {
 
         this.canvas = null;
         this.context = null;
+        this.initialCropArea = {
+            top: props.height || 100,
+            bottom: 0,
+            left: props.width || 100,
+            right: 0
+        };
         this.state = {
             width: props.width || 100,
             height: props.height || 100,
@@ -19,7 +25,8 @@ export default class SignatureInput extends preact.Component {
             isDrawing: false,
             lastX: 0,
             lastY: 0,
-            canvasImageExtractionBlocked: false
+            canvasImageExtractionBlocked: false,
+            cropArea: this.initialCropArea
         };
 
         this.handleMouse = this.handleMouse.bind(this);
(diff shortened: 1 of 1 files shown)


src/Utility/getCroppedCanvas.js
```diff
@@ -0,0 +1,78 @@
+export function getCroppedCanvas(ctx) {
+    const {width, height } = ctx.canvas;
+    if (!width && !height) { return ctx.canvas; }
+
+    const imgData = ctx.getImageData(0, 0, width, height);
+    const data = new Uint32Array(imgData.data.buffer);
+    data[0] = 0; // TODO: pointerout handler colors pixel at 0 0 black, so we set it back to transparent;
+                 // I tracked the issue to detectBlockedCanvasImageExtraction function;
```
zner0L: Puh, sorry for leaving you to such badly documented code. I think `detectBlockedCanvasImageExtraction` might be responsible, but that's @baltpeter's code, so I'm pinging him.

src/Utility/getCroppedCanvas.js
```diff
@@ -0,0 +1,78 @@
...
+                bottom = height - y - 1; 
+                if (top) {
+                    shouldBreak = true; 
+                    break;
+                }
+            }
+        }
+
+        if (y > height - y && !top && !bottom) { return ctx.canvas; } // image is blank
+    }
```
zner0L: Have you looked into the performance of that? It seems rather performance heavy and I think I'd rather not crop the canvas than stop the webpage every time a user draws a signature. I was thinking that you might be able to extract the furthest drawing point by looking at the maximum/minimum x and y during drawing. That saves you a lot of time searching for the drawing by looping through the whole image.

zner0L:
@merelj Thank you for your contribution for quite a hard problem! Ok, so I think I don't quite understand my own code here: when you just used `value`, what problems did you run into? Was the canvas constantly updating or only once the pointer left it? Because I think that might be alright.

If the problems were more severe, you'll have to change quite a lot, even cross repo. The signatures are handled in [`datenanfragen/letter-generator`](https://github.com/datenanfragen/letter-generator/blob/master/Letter.js), more specifically at `Letter::handleSignature`. You'll need to include the `croppedValue` there and change all the default values. So I would suggested, you try making `value` work, maybe by controling the state via a new `stateValue`.

src/Utility/getCroppedCanvas.js
baltpeter: We were indeed not correctly restoring the old pixel. Thanks for pointing that out, @merelj. I opened #137 to fix that.

merelj:
@zner0L how about this? I used approach you mentioned in your comment (now it is requires a bit more work on every mouse event, but there is no need to run through every pixel for cropping)

there's still a problem with the cropped image getting scaled up though...

my guess is it has to do with [this line](https://github.com/datenanfragen/letter-generator/blob/master/Letter.js#L84) in repo you mentioned earlier since image does not get scaled if its width matches the ope specified there.
image
is there a reason for specifying width there?

src/Utility/getCroppedCanvas.js
merelj: thank you for fixing it so quickly 👍

zner0L:
> is there a reason for specifying width there?

You were right, there is none apparently. I opened datenanfragen/letter-generator#3 to address this.

src/Components/Generator/SignatureInput.js
```diff
@@ -61,8 +68,10 @@ export default class SignatureInput extends preact.Component {
             // see https://stackoverflow.com/a/4776378
             let img = new Image();
             img.onload = () => {
+                const left = (this.canvas.width - img.width) / 2;
+                const top = (this.canvas.height - img.height) / 2;
```
zner0L: Since you are already working with the image, please also set the `cropArea` to the area of the image. Otherwise the image will just be cropped away if the user doesn't draw a circle around it (defeating the purpose of prefilling).

zner0L:
Also, I don't know if you noticed, but you can get extra swag from our own Hacktoberfest program for your contribution. Check out this [blog post](https://www.datarequests.org/blog/hacktoberfest-2019/) to learn more!

merelj:
@zner0L  fixed it, thanks for your hint :)

it's cool news about the swag, thanks for sharing, not sure I'm gonna be on time but I absolutely don't mind working on this stuff just to learn something useful 😅

zner0L:
> it's cool news about the swag, thanks for sharing, not sure I'm gonna be on time but I absolutely don't mind working on this stuff just to learn something useful 😅

@merelj Your PR will be merged eventually, so that definitely was opened on time. If you sign up for our program today you can still get the swag. Tomorrow the deadline is over, though.

PR #139, opened 2019-10-16
Tutorial at generator page now displaying the right position depending on the screen width.

### LINKED ISSUE #136: Tutorial at generator page step 1/4
At the generator page when it's showing the tutorial

image

It says on the right but the box is on the left.
I found the file to facilitate the fix. (good first issue for someone)

**website/src/i18n/en.json**
Line 319: `"request-parameters": "<p>Alright, let’s start generating your requests. We will try our best to make this as easy as possible for you.</p><p>On the right, we have already prefilled the contact details for the first company.</p>"`

I think it should have a class to show different strings
One for On the Left (Desktop) and one for On the Top (Mobile / Tablet)
Something like

`<p class="desktop">On the left, we have already prefilled the contact details for the first company.</p><p class="mobile">On top, we have already prefilled the contact details for the first company.</p>`

luko0610:
Tutorial at generator page now displaying the right position (On top, On the left) depending on the screen width. 
The problem has been fixed for the English and the German site.

Closes #136


src/styles/generator.scss
```diff
@@ -211,3 +211,26 @@ h2 {
     margin-top: 10px;
     margin-bottom: 10px;
 }
+
+/*Displaying wizard text based on screen size*/
+.p-wizard-desktop {
+    display: none;
+}
+
+@media screen and (max-width: 980px) {
```
baltpeter: Initially, I wanted to suggest using our [`mobile` mixin](https://github.com/datenanfragen/website/blob/master/src/styles/variables.scss#L47).

But then I noticed (and remembered…) that we don't even need that. We already have classes `.mobile-only` and `.desktop-only` ([here](https://github.com/datenanfragen/website/blob/master/src/styles/framework.scss#L107)) that do the exact same thing.

Let's just use those and simplify the PR.

luko0610: Okay, thanks for your feedback. I'll fix that asap.

luko0610:
Okay, I changed it. Now it is using the already existing css classes.

PR #140, opened 2019-10-16
Tutorial at generator page now displaying the right position depending on the screen width

luko0610:
By using the already existing css classes mobile-only and desktop-only.


baltpeter:
Thanks!

PR #157, opened 2019-10-22
Better UI for 'Send mail' buttons 

baltpeter:


src/Components/Generator/ActionButton.js
```diff
@@ -5,8 +5,9 @@ import MailtoDropdown from '../MailtoDropdown';
 
 export default class ActionButton extends preact.Component {
     render(props) {
+        const enabled = props.transport_medium === 'email' ? !!props.letter : props.download_active;
```
zner0L: If you create that `const` you can just as well use it further down.

baltpeter: Done.

src/Components/MailtoDropdown.js
```diff
@@ -0,0 +1,99 @@
+import preact from 'preact';
+import { Text, IntlProvider } from 'preact-i18n';
+import PropTypes from 'prop-types';
+import t, { t_r } from '../Utility/i18n';
+
+// TODO: I would like to extend this with country-specific services in the future. Unfortunately, I did not find a way
+// to produce links for GMX and Web.de.
+export const MAILTO_HANDLERS = {
+    mailto: {
+        link: d => `mailto:${d.email}?subject=${d.subject}&body=${d.body}`,
```
zner0L: While I was resolving the merge conflicts I realized that there is another, now useless method `RequestLetter::toMailtoLink()`. That should be removed if we decide to handle all this here.

baltpeter: We indeed don't seem to use that anywhere else.

Issue #154, opened 2019-10-22
Bug: Changing the language


nishantwrp:
The confirmation box appears two times when user tries to switch the language

baltpeter:
Thanks for the report.

PR #162, opened 2019-10-22
Fixed Bug: Changing the language #154

sohanpal:
There were 2 event listeners which causes to open modal 2 times.

### DIFF
--- a/src/Components/I18nWidget.js
+++ b/src/Components/I18nWidget.js
@@ -66,10 +66,7 @@ export default class I18nWidget extends preact.Component {
                             <Text id="language" />
                         </h2>
                         <div className="select-container">
-                            <select
-                                value={LOCALE}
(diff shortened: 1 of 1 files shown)


baltpeter:
Thanks for the PR!

@zner0L I believe there was a reason we added both event listeners. Do you remember why we did that?

zner0L:
Yes, that is for a11y reasons. I don't remember the specifics but we might want that. Can you look into that @baltpeter?

zner0L:
@baltpeter ?

zner0L:
Actually, I think this is fine. `onBlur` is the important event and that is still implemented. Sorry it took so long…

zner0L:
This is, however, a little counter intuitive because now you have to select the language and then click anywhere to blur the select. Still better than two popups, though. Smarter solutions are very welcome!

PR #169, opened 2019-10-24
Fixes #65: Objections

### LINKED ISSUE #65: Right to object
I think it would be a valuable addition to the site to generate "Right to object" letters.

Often we interact with companies that for various reasons we want to have and process our data, however by default these companies also process our data for the purpose of things like direct marketing etc.

According to the GDPR we are able to object to the processing of this data, but how to do so isn't always clear.

For example, the Austrian bank Sparkasse, [has a detailed section](https://www.sparkasse.at/sgruppe/wir-ueber-uns/datenschutz-sicherheit) on how they process our data and they state that we can object, but as a lay-person I'm not sure how to effectively formulate a letter that states I object to their use of my data for particular purposes.

baltpeter:


src/Components/Generator/RequestTypeChooser.js
```diff
@@ -2,7 +2,7 @@ import preact from 'preact';
 import t from 'Utility/i18n';
 import Radio from '../Radio';
 
-const VALID_REQUEST_TYPES = ['access', 'erasure', 'rectification', 'custom'];
+const VALID_REQUEST_TYPES = ['access', 'erasure', 'rectification', 'objection', 'custom'];
```
zner0L: Let's abstract this somewhere and also use it in here https://github.com/datenanfragen/website/pull/169/files#diff-038cd559a6195a66a3d95ed073da4eadR54

src/i18n/en.json
```diff
@@ -191,6 +192,7 @@
         "request-type": "What kind of request do you want to generate?",
         "access-request": "Access request",
         "erasure-request": "Erasure request",
+        "objection-request": "Objection against direct marketing",
```
zner0L: This is pretty long and looks ugly on mobile devices. What about "Direct marketing objection"?

baltpeter: Feels like a bit of a "German-ism" to me but sure.

PR #172, opened 2019-10-27
Improve Code Looks

tokirmanva22:
remove extra things.


baltpeter:
Sorry but those are not changes we want.

The curly braces are important to us, without them it becomes a lot easier to make errors. Also, we *much* prefer the look of those 'on-line if statements'.

Besides, we are already using Prettier for code formatting.

Issue #156, opened 2019-10-22
id-data heading padding


davenewham:
The formatting on the Rectification request input field on the right handside for "My data" heading seems a little off. Seems like the correct padding is needed after the input field?

It is more prevelant in non-mobile resolutions. 

![grafik](https://user-images.githubusercontent.com/15076177/67279336-14710380-f4cb-11e9-8983-0d792ff28087.png)

baltpeter:
Thanks for the report!

Should be easy enough to fix, so I'll leave that for someone else.

parammittal16:
@baltpeter can I do this ?

baltpeter:
This was fixed in #316.

PR #178, opened 2019-10-31
Heading margin

davenewham:
Attempt at fixing #156. The fallthrough of the switch statement was interesting 🤔


src/Components/Generator/DynamicInputContainer.js
```diff
@@ -96,7 +96,7 @@ export default class DynamicInputContainer extends preact.Component {
         return (
             <IntlProvider scope="generator" definition={I18N_DEFINITION}>
                 <div className="dynamic-input-container">
-                    <h2>{this.props.title}</h2>
+                    <h2 style={this.props.margin}>{this.props.title}</h2>
```
zner0L: I think this should rather be solved by adding a class on a flag. The name of the attribute here is misleading (`heading-syle` would be more appropriate) and `margin-top: 0` is already set via a more general style. Simply add a class `has-margin` to the heading and then add the style to the SCSS at `.box h2.has-margin`.

zner0L:
Sorry for taking so long for reviewing your work. I must have missed it in between my mails!

zner0L:
Since there has been no feedback for quite a while an this PR is about a trivial change, I will permanently close it. Switching to the new codebase will be more work than just implementing again.

Issue #173, opened 2019-10-28
Fix RSS support for Datenanfragen.de blog


m-rey:
Currently, https://www.datenanfragen.de/blog/ has no RSS support (afaik).
A rss/atom feed is an easy way to stay updated using a feed reader.

baltpeter:
I'm definitely a proponent of RSS as well. We do have an RSS feed at: https://www.datenanfragen.de/index.xml

While not exposed in the UI anywhere, it is specified in the meta tags:

```html
<link href=https://www.datenanfragen.de/index.xml rel=alternate type=application/rss+xml title=Datenanfragen.de>
<link href=https://www.datenanfragen.de/index.xml rel=feed type=application/rss+xml title=Datenanfragen.de>
```

I believe most feed readers should recognize that if you just enter the URL of the blog (or even the homepage for that matter).

baltpeter:
Actually, on further inspection: That feed is empty.

However https://www.datenanfragen.de/blog/index.xml does seem to have the blog posts.

So, there's definitely a bug here that we'll have to investigate. Thanks for reporting!

HoffmannP:
So there are two possible solutions. The one will link to the index.xml of each folder as `.RSSLink` might have done. Might, b/c `.RssLink` is [deprecated](https://gohugo.io/variables/page/) now.

But I think you'd rather want to link to the one RSS-feed for blog articles. That's what I've changed the current version to.

I post the alternative for completeness:

    <link href="{{ with .OutputFormats.Get "RSS" }}{{ .Permalink }}{{ end }}" rel="feed" type="application/rss+xml" title="{{ .Site.Title }}" />

Attached you find the difference for several pages:

```bash
index.html
48:<link href="https://www.datenanfragen.de/blog/index.xml" rel="alternate" type="application/rss+xml" title="Datenanfragen.de" />
49:<link href="https://www.datenanfragen.de/index.xml" rel="feed" type="application/rss+xml" title="Datenanfragen.de" />

blog/index.html
48:<link href="https://www.datenanfragen.de/blog/index.xml" rel="alternate" type="application/rss+xml" title="Datenanfragen.de" />
49:<link href="https://www.datenanfragen.de/blog/index.xml" rel="feed" type="application/rss+xml" title="Datenanfragen.de" />

blog/hacktoberfest-2019/index.html
52:<link href="https://www.datenanfragen.de/blog/index.xml" rel="alternate" type="application/rss+xml" title="Datenanfragen.de" />
[...]

PR #180, opened 2019-10-31
Fix RSS support for Datenanfragen.de blog #173

HoffmannP:
See Issue #173

### DIFF
--- a/layouts/partials/head/meta.html
+++ b/layouts/partials/head/meta.html
@@ -27,8 +27,8 @@
 {{ end }}
 
 
-<link href="{{ .Site.BaseURL }}blog/index.xml" rel="alternate" type="application/rss+xml" title="{{ .Site.Title }}" />
-<link href="{{ .Site.BaseURL }}blog/index.xml" rel="feed" type="application/rss+xml" title="{{ .Site.Title }}" />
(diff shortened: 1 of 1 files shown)


layouts/partials/head/meta.html
```diff
@@ -27,10 +27,8 @@
 {{ end }}
 
 
-{{ if .Site.RSSLink -}}
-<link href="{{ .Site.RSSLink }}" rel="alternate" type="application/rss+xml" title="{{ .Site.Title }}" />
-<link href="{{ .Site.RSSLink }}" rel="feed" type="application/rss+xml" title="{{ .Site.Title }}" />
-{{- end }}
+<link href="{{ .Site.BaseURL }}blog/index.xml" rel="alternate" type="application/rss+xml" title="{{ .Site.Title }}" />
+<link href="{{ .Site.BaseURL }}blog/index.xml" rel="feed" type="application/rss+xml" title="{{ .Site.Title }}" />
```
baltpeter: Usually, we would use `{{ "blog/index.xml" | absURL }}` instead of constructing the URL ourselves from the base URL.

I know they amount to the same thing but I would prefer to keep this consistent unless this approach doesn't work here for some reason.

HoffmannP: should be ok now

PR #192, opened 2019-11-14
Fixes #190: More inutitive use of the language chooser

### LINKED ISSUE #190: More inutitive use of the language chooser
This is, however, a little counter intuitive because now you have to select the language and then click anywhere to blur the select. Still better than two popups, though. Smarter solutions are very welcome!

_Originally posted by @zner0L in https://github.com/datenanfragen/website/pull/162#issuecomment-553987351_

For this to work I guess we have to reimplement `onChange` and do something smart to prevent the modal on opening twice if `onBlur` is triggered.

baltpeter:
Wouldn't exactly call it a 'smart' solution but it does the trick.

### DIFF
--- a/src/Components/I18nWidget.js
+++ b/src/Components/I18nWidget.js
@@ -15,12 +15,16 @@ export default class I18nWidget extends preact.Component {
         });
     }
 
+    static alreadyShowingModal = false;
     static showLanguageChangeModal(event) {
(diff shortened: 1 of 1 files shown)


zner0L:
While testing I noticed another bug: If the select is blurred without any change, the "Change language" button does not work.

PR #197, opened 2019-11-15
Fixes #193: Don't fail if the browser doesn't report a country

### LINKED ISSUE #193: Kann Land nicht festlegen
Im Auswahlfenster wird mir kein Land angezeigt und unten auf der Seite kann ich das Land auch nicht ändern.
grafik
grafik

baltpeter:


pyfisch:
:+1: Funktioniert alles wie erwartet.

src/general.js
```diff
@@ -78,7 +78,7 @@ function guessUserCountry() {
```
zner0L: We should also try to fallback to the language if we can not get the country from the language string, because only a language is reported (e.g. `de` instead of `de-de` as seen in #193).

PR #209, opened 2019-12-06
Merge verein-website into website

baltpeter:


baltpeter:
@zner0L This is ready for review now.

This needs to be merged first, followed by datenanfragen/verein-website#10.

content/open-source.de.md
```diff
@@ -90,12 +90,15 @@ Ein Teil der Entwicklungsarbeit findet momentan über eine private GitLab-Instan
     <a class="button button-primary read-more-button" href="https://github.com/baltpeter/yace">Zum Repository&nbsp;<span class="icon icon-arrow-right"></span></a>
 </article>
 
+<a id="license-notices"></a>
```
zner0L: hugo does provide anchors on it's own already.

content/verein/constitution.de.md
```diff
@@ -0,0 +1,174 @@
+{
+    "title": "Satzung",
+    "slug": "satzung",
+    "aliases": ["constitution"],
+    "type": "page"
+}
+
```
zner0L: Since it is now not completely clear what the constitution is for for every user, I think we need some kind of introduction here.

src/i18n/de.json
```diff
@@ -129,7 +129,70 @@
         "verein-home-card-donate-button": "Ich möchte spenden.",
         "verein-home-card-join-heading": "Uns beitreten.",
         "verein-home-card-join-content": "Unser Projekt lebt von Zusammenarbeit. Wenn Du in der Entwicklung und Verwaltung mitmachen möchtest oder einfach mehr von uns mitbekommen möchtest, heißen wir Dich herzlich bei uns willkommen.",
-        "verein-home-card-join-button": "Ich möchte beitreten."
+        "verein-home-card-join-button": "Ich möchte beitreten.",
+        "footer-tab-website-name": "Datenanfragen.de",
+        "footer-request-generator": "Anfrage generieren",
+        "footer-request-generator-url": "/generator",
```
zner0L: Why are the URLs even i18nable if the menu is hardcoded anyway? Shouldn't simply hardcoding it all do the trick?

src/Utility/common.js
```diff
@@ -56,6 +56,23 @@ export function hash(s) {
...
+        ('0' + d.getUTCDate()).slice(-2) +
+        Math.random()
+            .toString(36)
+            .substr(2, length)
+            .toUpperCase()
+    );
+}
+
+export function renderMoney(amount, currency = '') {
+    return Number(amount).toLocaleString(LOCALE, { minimumFractionDigits: 2 }) + (currency ? ' ' + currency : '');
```
zner0L: Apparently this shows 3 digits which is one too many…

zner0L: And the QR Codes seem to be broken in print.

content/open-source.de.md
baltpeter: Yeah but isn't that some ugly thing? I'd much rather have a clear one that I control.

content/verein/constitution.de.md
baltpeter: Not quite sure what you mean. What has changed? I mean, if you want to write an introduction, feel free but I don't see the need.

src/i18n/de.json
baltpeter: Nope. Two cases where this doesn't work:

* Some (most?) of the URLs are translated. `/donate` might redirect to `/spenden` but that redirect is ugly and unnecessary.
* For the French version (and potentially others in the future), we need to be able to link to the English sites that aren't translated yet.

src/Utility/common.js
baltpeter: > Apparently this shows 3 digits which is one too many…

Where? Looks fine for me.

baltpeter: > And the QR Codes seem to be broken in print.

Fixed.

baltpeter:
Note: Once this is merged, we should inform the ITZ of the new URL for the transparency page.

src/Utility/common.js
zner0L: > Where? Looks fine for me.

Screen Shot 2019-12-10 at 14 22 00

content/verein/constitution.de.md
zner0L: Before, the constitution was only accessible from the `verein.` webpage, not the general one. So one might stumble upon the constitution without knowing that daterequests.org etc. are run by an association.

But oh well, I will write a short intro.

src/Utility/common.js
baltpeter: 155d14a

content/verein/constitution.de.md
baltpeter: Well yeah, my point is that it was really easy to get there accidentally without realizing. Previously, you probably also clicked a link somewhere and where taken to a page (that to most user) looks exactly like the main page, so… But anyway.

PR #222, opened 2020-02-24
Translate "Donate" page into French

NathanBnm:


NathanBnm:
Is the `type` attribute used in the URL?

content/donate.fr.md
```diff
@@ -0,0 +1,58 @@
+{
+  "title": "Faire un don",
+  "type": "faire-un-don",
```
baltpeter: Sorry, this is not exactly obvious: The type doesn't influence the URL but does actually specify what kind of page this is. Usually, this would just be `page` but here we need the type `donate` because there is a special layout that will be pulled in and include the donation form (i.e. fill the `<div class="donation-widget"></div>` through JavaScript).

content/donate.fr.md
```diff
@@ -0,0 +1,58 @@
...
+  "aliases": ["spenden", "verein/donate", "verein/spenden"]
+}
+
+Merci beaucoup de ton soutien ! Ton don nous aidera à financer nos projets actuels et futurs afin qu'ils restent gratuits pour tous.
+
+Si tu souhaites nous soutenir de manière récurrente, tu peux également [devenir membre]({{< ref "become-a-member" >}}). De cette façon, tu peux également nous fournir un mandat de prélèvement SEPA.
+
+En tant qu'organisation d'utilité publique, nous pouvons également te délivrer un [reçu de don](#donation-receipt-howto).
```
baltpeter: I am assuming that French tax authorities won't accept donations to us for tax deductions. As such, this sentence is probably more confusing than helping.

Let's just delete this sentence here and leave the explanation below.

content/donate.fr.md
```diff
@@ -0,0 +1,58 @@
...
+* **Comment puis-je donner de façon anonyme ?**
+
+  Si tu ne souhaites pas indiquer ton identité, tu peux soit utiliser une monnaie cryptée anonyme, soit nous envoyer de l'argent liquide par courrier à notre adresse :
+
+    *Datenanfragen.de e.&thinsp;V.  
+    Schreinerweg 6  
+    38126 Braunschweig  
+    Allemagne*
+
+  Pour les dons en espèces, nous te conseillons d'utiliser le [WERT NATIONAL](https://www.deutschepost.de/de/w/wert-national.html) du service postal allemand ou un produit similaire proposé par ton bureau de poste local. Nous ne recommandons pas d'envoyer des dons en espèces à l'étranger. Dans tous les cas, assure-toi que le contenu de ta lettre n'est pas visible de l'extérieur.
```
baltpeter: "Wert National" is an insured and officially sanctioned way of sending cash by mail in German (you usually aren't supposed to). From a quick glance through the La Poste website, they don't seem to offer a similar product.

Unless you know of one, I think we can just get rid of the reference to "Wert National".

content/donate.fr.md
NathanBnm: I was not sure that's why I asked. I'll fix that.

content/donate.fr.md
NathanBnm: Actually I think you're wrong. "The body may be in France or a Member State of the **European Union**, Iceland, Norway or Liechtenstein" as written here https://www.service-public.fr/particuliers/vosdroits/F426 (French only)

baltpeter: Oh nevermind. Good to know, thanks for checking! In that case, this sentence can of course stay.

content/donate.fr.md
NathanBnm: You're right, in France there is no guarantee on it and no postal service to send cash. I'll remove that.

NathanBnm:
> 
> 
> Thanks a lot! Just a couple small requests for changes.
> 
> Currently, the build fails due to undefined references. Once you've made the changes, I'll combine all PRs into one and remove the remaining undefined references, so we can get this merged.

That's fine, do not hesitate to notice me if I do something wrong :wink:

Btw I resolved your comments on this PR.

baltpeter:
Combined into #237.

PR #223, opened 2020-02-24
Translate "Thank you for your donation!" page into French

NathanBnm:
I don't know if I should translate the e-mail subject and body too.


baltpeter:
Combined into #237.

PR #227, opened 2020-02-24
Translate "Become a supporting member" into French

NathanBnm:


NathanBnm:
>  You are working faster than I can review. Awesome! :D

Btw I have trouble to answer to your reviews too x)

PR #230, opened 2020-02-25
Translate "The board" page into French

NathanBnm:


NathanBnm:
I was thinking. Can you and Lorenz be told as "Associated chairman"? If it's the same it would be better for the French translation as "Président associé" which implies that there are more than one chairman.

baltpeter:
I don't think I've heard of the term "Associated chairman" before, so I can't really say if it's correct. I'll try to explain a little: Our constitution does require that there always be exactly two chairpersons. We both represent the association jointly, i.e. there is no first and second chairman but instead we both have equal 'rights' with regards to the representation of the association. Usually, decisions are made jointly by the both of us (although in some cases decisions can be delegated to a single chairman).

NathanBnm:
I think it could be appropriate. That's what I had in mind. I will check and go back to you.

baltpeter:
Combined into #237.

PR #234, opened 2020-02-25
Translate "Transparency" page into French

NathanBnm:
Will this page be updated with 2019 information?

I think there is a typo in the original file at line 48. I think there should be only one f instead of two.

### DIFF
--- a/content/verein/transparency.fr.md
+++ b/content/verein/transparency.fr.md
@@ -6,7 +6,7 @@
 
 <a href="https://www.transparency.de/mitmachen/initiative-transparente-zivilgesellschaft/" class="no-link-decoration"><img src="/img/logo-itz.svg" style="float: right; width: 40%; min-width: 200px; padding: 5px;"></a>
 
-Afin que vous puissiez nous faire confiance, nous nous sommes engagés à être aussi transparents que possible sur nos procédures, nos finances et nos décisions. C'est pourquoi nous avons adhéré à l'[Initiative for a Transparent Civil Society](https://www.transparency.de/mitmachen/initiative-transparente-zivilgesellschaft/), une initiative de la branche allemande de Transparency International, qui est autorisée à nous contrôler par échantillonnage. Ils garantissent le respect de notre promesse de gestion transparente. Sur cette page, nous avons résumé toutes les informations importantes que nous publions dans le cadre de cette initiative.
+Afin que vous puissiez nous faire confiance, nous nous sommes engagés à être aussi transparents que possible sur nos procédures, nos finances et nos décisions. C'est pourquoi nous avons adhéré à l'[Initiative pour une Société Civile Transparente](https://www.transparency.de/mitmachen/initiative-transparente-zivilgesellschaft/), une initiative de la branche allemande de Transparency International, qui est autorisée à nous contrôler par échantillonnage. Ils garantissent le respect de notre promesse de gestion transparente. Sur cette page, nous avons résumé toutes les informations importantes que nous publions dans le cadre de cette initiative.
 
 Ces informations ont été mises à jour pour la dernière fois le 1<sup>er</sup> août 2019.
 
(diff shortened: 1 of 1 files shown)


baltpeter:
> Will this page be updated with 2019 information?

Absolutely. We will try to finish our yearly report for 2019 as quickly as possible and will then also update the transparency page. If you want, I can tag you in the PR for that update once it's there.

> I think there is a typo in the original file at line 48. I think there should be only one f instead of two.

In German, you use 'f' for 'folgende' and 'ff' for 'fortfolgende'*, meaning not just the next page but the next couple of pages. This might very well have just been copied from the German version and I also wasn't sure if this is also done in English but [Wiktionary](https://en.wiktionary.org/wiki/ff) does agree with our use here.

---
* Fun fact from looking at the [German Wiktionary page](https://de.wiktionary.org/wiki/ff.): Apparently, I am wrong and this is a common misconception. The second 'f' just signifies the plural and the word 'fortfolgende' is invented and doesn't actually exist. Go figure, the more you know. :D

content/verein/transparency.fr.md
```diff
@@ -0,0 +1,84 @@
+{
+    "title": "Transparence",
+    "slug": "transparence",
+    "type": "page"
+}
+
+<a href="https://www.transparency.de/mitmachen/initiative-transparente-zivilgesellschaft/" class="no-link-decoration"><img src="/img/logo-itz.svg" style="float: right; width: 40%; min-width: 200px; padding: 5px;"></a>
+
+Afin que vous puissiez nous faire confiance, nous nous sommes engagés à être aussi transparents que possible sur nos procédures, nos finances et nos décisions. C'est pourquoi nous avons adhéré à l'[Initiative for a Transparent Civil Society](https://www.transparency.de/mitmachen/initiative-transparente-zivilgesellschaft/), une initiative de la branche allemande de Transparency International, qui est autorisée à nous contrôler par échantillonnage. Ils garantissent le respect de notre promesse de gestion transparente. Sur cette page, nous avons résumé toutes les informations importantes que nous publions dans le cadre de cette initiative.
```
baltpeter: As far as I remember, "Initiative for a Transparent Civil Society" is just a translation by us. I think the only official name is the German "Initiative Transparente Zivilgesellschaft". So, do feel free to translate that name into French for easier readability.

NathanBnm:
> If you want, I can tag you in the PR for that update once it's there.

Yes for sure!

> In German, you use 'f' for 'folgende' and 'ff' for 'fortfolgende'*, meaning not just the next page but the next couple of pages. This might very well have just been copied from the German version and I also wasn't sure if this is also done in English but [Wiktionary](https://en.wiktionary.org/wiki/ff) does agree with our use here.
> 
> Fun fact from looking at the [German Wiktionary page](https://de.wiktionary.org/wiki/ff.): Apparently, I am wrong and this is a common misconception. The second 'f' just signifies the plural and the word 'fortfolgende' is invented and doesn't actually exist. Go figure, the more you know. :D

Oh I didn't know that, thanks for explanation. In fact this symbol does not exist in French :smile:

content/verein/transparency.fr.md
```diff
@@ -0,0 +1,84 @@
...
+
+Nous publions la constitution complète de notre association sur [ce site]({{< ref "verein/constitution" >}}) et sur [GitHub](https://github.com/datenanfragen/verein/blob/master/satzung.md).
+
+Notre objectif est le même que celui énoncé dans notre constitution, mais nous l'avons également résumé et expliqué [ici]({{< ref "verein/mission-statement" >}}).  
+
+## 3. Informations sur les avantages fiscaux
+
+Le centre des impôts de Braunschweig-Wilhelmstraße nous a agréés en tant qu'association d'intérêt général par la [déclaration du 21 juin 2018 (en allemand)](https://static.dacdn.de/docs/feststellungsbescheid_2018-06-21.pdf) pour la réalisation d'activités de conseil et de protection ainsi que pour la promotion du gouvernement démocratique. Le numéro d'identification fiscale suivant nous a été attribué : 14/209/11640
+
+## 4. Name and function of significant decision makers
```
baltpeter: You missed this line. :)

content/verein/transparency.fr.md
```diff
@@ -0,0 +1,84 @@
...
+
+L'association est gérée par le [comité directeur]({{< ref "verein/board" >}}). Elle est composée de :
+
+ - Benjamin Altpeter, président associé
```
baltpeter: I am nitpicking here but let's have both 'P's either capitalized or lowercase.

baltpeter:
Combined into #237.

PR #235, opened 2020-02-25
Translate "Constitution" page into French

NathanBnm:


content/verein/constitution.fr.md
```diff
@@ -0,0 +1,181 @@
...
+d) la sensibilisation du public et la réalisation de campagnes, ainsi que l'organisation ou la participation à des conférences, séminaires et autres événements destinés à conseiller le grand public sur le droit à la vie privée et à faire avancer le débat concernant un droit fondamental à la liberté d'information, et  
+e) la gestion et le développement des plates-formes en ligne ouvertes, gratuites et accessibles au public datenanfragen.de (en allemand), datarequests.org (en anglais), demandetesdonnees.fr (en français) et éventuellement dans d'autres langues, qui permettent d'accéder au générateur décrit au point a) et à la base de données décrite au point c), de rendre le matériel d'information décrit au point b) accessible au public et, enfin, de faire de la publicité pour la sensibilisation du public décrite au point d).
```
baltpeter: Two small remarks here:

* The 'd' in 'Datenanfragen.de' should be capitalized.
* Usually, you are absolutely right with including demandetesdonnees.fr in these lists. In this case however, as this is a translation of an official legal document which doesn't mention dtd.fr, I would rather leave it out.

content/verein/constitution.fr.md
```diff
@@ -0,0 +1,181 @@
...
+    "slug": "constitution",
+    "aliases": ["statutes", "statute"],
+    "type": "page"
+}
+
+Notre constitution fixe le cadre juridique de l'association Datenanfragen.de e.&nbsp;V. Elle explique entre autres nos objectifs et la manière dont nous voulons les atteindre, elle fixe les règles concernant l'adhésion à notre association et définit le processus décisionnel.
+
+<p class="box box-info" style="margin-bottom: 20px;">
+Comme nous sommes basés en Allemagne, seule la <a href="https://www.datenanfragen.de/verein/satzung">version allemande</a> de notre constitution est juridiquement applicable. Cette traduction n'est proposée que par souci de confort.<br>
+Dans les cas où nous pensons que la traduction française peut être ambiguë, nous avons fait référence au texte original allemand [entre crochets] pour plus de clarté.
```
baltpeter: Seems like it is easier to convey German legal concepts in French than it is in English. :D As you didn't make use of any clarifications like this, we can remove this sentence.

content/verein/constitution.fr.md
NathanBnm: Even as an URL? I understand it should be as the association name but generally URLs should not contain capital letters.
That's okay for the second point :wink:

content/verein/constitution.fr.md
NathanBnm: You're right. I forgot to remove it. When I was not sure I checked the german version too.
I still make my translations with the help of DeepL and it seems that it is quite good at translating "legal" sentence style. Because it's based on previously translated texts and AI it's really accurate.

content/verein/constitution.fr.md
baltpeter: I agree that URLs don't usually contain capital letters but in this case this *is* the name of the website that just happens to look like (and be the same) as the website URL. :D That's just the 'style guide' we have defined.

If however we were to write an actual URL that isn't just the name (like www.datenanfragen.de/generator), the 'd' would indeed be lower-case. Although this usually only happens in print documents. On the web, we of course use actual links.

NathanBnm: Understood 👍

baltpeter:
Combined into #237.

PR #240, opened 2020-02-27
Fix "Thanks" page type

NathanBnm:


NathanBnm:
It wasn't because it was the same suggestion that the one with the slug!

PR #243, opened 2020-02-27
Rename "Privacy controls" page in French

NathanBnm:
Already done via Weblate but forgot to do so here.

### DIFF
--- a/content/privacy-controls.fr.md
+++ b/content/privacy-controls.fr.md
@@ -1,6 +1,9 @@
 {
     "title": "Options de confidentialité",
     "slug": "options-de-confidentialite",
+    "aliases": [
+        "privacy-controls"
+    ],
     "type": "privacy-controls",
     "static": true
 }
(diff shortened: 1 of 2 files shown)


NathanBnm:
What do you mean exactly? Why don't we use the same rule everywhere.

baltpeter:
> What do you mean exactly? Why don't we use the same rule everywhere.

Not quite sure what you are referring to.

The privacy controls page can have a translated slug but in addition to that, it also needs to be reachable via `/privacy-controls`. That link is hardcoded in various places that aren't easily translatable.

The footer is also currently linking to `/privacy-controls`. That should be changed to the new URL.

NathanBnm:
@baltpeter we can simply translate the URL like in #264

baltpeter:
The problem is that what I did in #264 only works in Hugo templates, not in JavaScript. And either way, I am worried that we will miss some usages of the hardcoded URL. But as I said, I am not opposed to translating the URL. I just want to also add an alias to prevent problems with old links.

NathanBnm:
> The problem is that what I did in #264 only works in Hugo templates, not in JavaScript.

I'm not familiar with Hugo so I wasn't aware about that 😅 

> And either way, I am worried that we will miss some usages of the hardcoded URL. But as I said, I am not opposed to translating the URL. I just want to also add an alias to prevent problems with old links.

Right, so let's do that!

PR #269, opened 2020-03-07
Restructure content directory

### LINKED ISSUE #85: Restructure content dir
We could structure our content more sensibly.

Suggestions:

* Use folders for i18n instead of filenames.
* Include images for a blog post in the corresponding content folder instead of in the `static` dir.

NathanBnm:
This PR closes #85.

@baltpeter suggestions were:

- [X] Use folders for i18n instead of filenames.
- [X] Include images for a blog post in the corresponding content folder instead of in the `static` dir.

I have build the website locally and everything is fine. I didn't include images for the blog posts in the content folder yet. I was wondering if we just have to create an `img` folder into the `blog` directory of each language. The fact is that if we do this, some images will be duplicated as they don't all need to be translated. Any proposition?

Finally I would like you to check the `deploy.sh` script. I've updated it but and added some execution log messages but it could maybe be better.

This PR is still a work in progress and every review and proposition is welcome :smile:


deploy.sh
```diff
@@ -1,44 +1,64 @@
 #!/usr/bin/env bash
 set -e
 
-# --- Fetch companies data ---
+echo "--- Fetching data ---"
```
baltpeter: Having the output is probably a good idea. Not sure if the `---` is the best fit though. I used those to let the comments 'stand out' from the code. Is that necessary in the terminal output? Maybe just a simple `Fetching data…` instead?

baltpeter:
> I didn't include images for the blog posts in the content folder yet. I was wondering if we just have to create an img folder into the blog directory of each language. The fact is that if we do this, some images will be duplicated as they don't all need to be translated. Any proposition?

That's a good point. Not sure to be honest. Currently the images are scattered all over the place in the `static` folder. Recently, I have tried to be more organized but it isn't exactly great yet. Having them sorted with their respective post would be quite an advantage, I think.

deploy.sh
```diff
@@ -1,44 +1,64 @@
...
 git clone --depth 1 https://github.com/datenanfragen/data data_tmp
-mkdir -p content/company
-mkdir -p content/supervisory-authority
+
+echo "--- Creating directories ---"
+for lang in de en fr
```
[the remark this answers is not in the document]
NathanBnm: I was thinking about it but couldn't manage to make it work 😅 I will try again

deploy.sh
NathanBnm: You're right, I just copy-pasted it 😄 I will change this.
I think it's fine to print these comments in the terminal for non-owners of the project who wants to build it locally like me.

NathanBnm:
>Having them sorted with their respective post would be quite an advantage, I think.

How would you manage to make the difference between common images and specific images? I mean as I said before some of them doesn't need to be translated and they will be duplicated as many times as there are languages.

baltpeter:
> How would you manage to make the difference between common images and specific images? I mean as I said before some of them doesn't need to be translated and they will be duplicated as many times as there are languages.

What about leaving the non-translated images in `static/img/blog` and only moving the translated images?

NathanBnm:
> What about leaving the non-translated images in `static/img/blog` and only moving the translated images?

That's what I was thinking about, but we have to establish a better documentation for translators in order to help them to do their work.

baltpeter:
> That's what I was thinking about, but we have to establish a better documentation for translators in order to help them to do their work.

This would only affect translating the Markdown posts directly, wouldn't it? That currently isn't documented at all, which is something I agree we should improve.

NathanBnm:
> This would only affect translating the Markdown posts directly, wouldn't it? That currently isn't documented at all, which is something I agree we should improve.

For sure! We have to explain which files have to be translated and how.

NathanBnm:
I'm gonna work on that as soon as I will have time. I'm pretty busy for now.

Could you just please notify it here when some changes the the `content` folder is applied (I'm thinking about #274) for me to apply changes here too :smile:

baltpeter:
Just a quick note: Thanks to @anikarini, we now have a Portuguese translation which I am currently in the process of deploying. Thus, `pt` will need to be added to `deploy.sh`.

baltpeter:
There is now a new German blog post: https://github.com/datenanfragen/website/blob/master/content/blog/wie-oft-anfragen.de.md

NathanBnm:
@baltpeter I moved some of the content from the `static` directory to the `content` directory. I have created `img` subfolders to avoid some files mess.

I was wondering if we could do the same for the `static/templates` and `static/downloads` content. What do you think?

I added your new german blog post and added pt language to `deploy.sh` script too.

And what about the `static/card` folder?

That's a lot of questions, but would it be possible to put the translated logos here too?

I was also wondering what to do with the `static/blog/online-privacy` content.

NathanBnm:
@baltpeter I have an issue when building the site but can't manage to see what's causing it. Any idea? You can check the CI logs.

baltpeter:
Thanks for the continued work.

> I have an issue when building the site but can't manage to see what's causing it. Any idea? You can check the CI logs.

That is indeed not a very helpful error message, sorry about that. The problem is that the new `wie-oft-anfragen` post has a new author (`nadjarode`) that isn't yet in your working copy.  
To avoid problems like this (there might be others that might just not be as obvious as this one), I would recommend rebasing against master rather than copying the new content manually. Merging is also ok, if you prefer that.

> I moved some of the content from the static directory to the content directory. I have created img subfolders to avoid some files mess.

Unfortunately, the images inserted into blog posts now break (not the featured ones). As you have removed the `/` at the start of the URLs, they are now relative to the post URL whereas the actual files are relative to `/blog`. I'm afraid to avoid over-complicating the Markdown files, we'll have to invent a shortcode similar to the [one the Hugo docs site uses](https://github.com/gohugoio/hugoDocs/blob/master/layouts/shortcodes/imgproc.html#L1-L15). We don't need the HTML generation, just figuring out the actual URL is enough. Maybe we can add the option to process the images since that is now possible due to the changes you made, but we can also add that later.

> I was wondering if we could do the same for the static/templates and static/downloads content. What do you think?

Considering that will change the URLs of those files, I would rather not. The `static/templates` stuff is not actually tied to any content but rather generated by the deploy script and used in the generator. And as for `static/downloads`, there are unfortunately already quite a few links from third-party sites to those. I definitely don't want to break those.  
We could either leave the current downloads as legacy problems and only apply the new scheme for future downloads. Or I guess we could also setup redirects…

> And what about the static/card folder?

Those are generic icons used in multiple places (including places not associated with any content).

> That's a lot of questions, but would it be possible to put the translated logos here too?

I guess that would work technically, I just don't think it makes sense semantically. Besides, I would also prefer not to break those URLs.

> I was also wondering what to do with the static/blog/online-privacy content.

Those are part of `content/en/blog/browser-extension-recommendations.md`.

NathanBnm:
@baltpeter thanks for your answer, I'm gonna check everything then. I'm gonna try to rebase the master branch too.

NathanBnm:
> To avoid problems like this (there might be others that might just not be as obvious as this one), I would recommend rebasing against master rather than copying the new content manually.

Done! :wink:

> Unfortunately, the images inserted into blog posts now break (not the featured ones). As you have removed the `/` at the start of the URLs, they are now relative to the post URL whereas the actual files are relative to `/blog`. I'm afraid to avoid over-complicating the Markdown files, we'll have to invent a shortcode similar to the [one the Hugo docs site uses](https://github.com/gohugoio/hugoDocs/blob/master/layouts/shortcodes/imgproc.html#L1-L15). We don't need the HTML generation, just figuring out the actual URL is enough. Maybe we can add the option to process the images since that is now possible due to the changes you made, but we can also add that later.

I have fixed that for now, but sure it would be great to do that. I just discovered a bug about featured images I'll explain it in another comment.

> We could either leave the current downloads as legacy problems and only apply the new scheme for future downloads. Or I guess we could also setup redirects…

Would be better to make everything following the same scheme. You can always add some redirects and maybe inform the sites (don't know how many) in order for them to update their link to the new one. And when every site has done it then redirect can be removed.

> I guess that would work technically, I just don't think it makes sense semantically. Besides, I would also prefer not to break those URLs.

About the logos, I think we should create a dedicated folder in the `static` folder like `logos` for more consistency.

> Those are part of `content/en/blog/browser-extension-recommendations.md`.

Moved!

NathanBnm:
I have the following issue with featured images:

image

They do not appear after page 1. Here is page 2.

The link of the image is actually `domain/blog/page/2/img/gdpr-erasure-request-letter.jpg` where it should be `domain/blog/img/gdpr-erasure-request-letter.jpg`.

On page 1 this is working correctly. I don't really know why this is doing that. Any idea?

NathanBnm:
I was wondering, what does the `act` folder stands for? And should it be translated?

Moreover what about moving `verein` folder to `association` for better consistency with localization? Personally I think it can look weird for translators to have everything in English and a German word for this folder.

It's not really the purpose of this PR so we can do this later but just asking.

Finally, are you going to decline your material in different languages? If it's the case maybe it has to be moved to the content directory too?

NathanBnm:
@baltpeter could you please check everything when you have time? :smile:

baltpeter:
Before I reply: Would it be ok if I force-pushed to your branch to try and cleanup the history of the branch a little? Reviewing is a little more difficult if the changes of the last couple of months are all in the diff. :D

NathanBnm:
@baltpeter sure you can

baltpeter:
> I have fixed that for now, but sure it would be great to do that. I just discovered a bug about featured images I'll explain it in another comment.

Thanks for changing the URLs manually. I think we should implement this shortcode before merging. If you're up for that, go for it, otherwise I can do that.

> Would be better to make everything following the same scheme. You can always add some redirects and maybe inform the sites (don't know how many) in order for them to update their link to the new one. And when every site has done it then redirect can be removed.

Ok. Let's move the downloads and setup redirects for the old URLs.

I don't think it's feasible to try and contact all linking sites (for starters we don't have a list of them). I don't mind keeping the redirects around anyway, though, so that's fine.

> About the logos, I think we should create a dedicated folder in the static folder like logos for more consistency.

As long as we also setup redirects for the remaining ones, that's fine with me. I remember using these URLs in other places before (like [here](https://github.com/datenanfragen/material-hlci-2020/blob/master/kurse/brand-identity.md)), so they need to keep working.

> I have the following issue with featured images:

Good catch! That's pretty much the same issue as with the images in the blog posts. The featured image URLs are now relative instead of absolute and the summary template doesn't know how to deal with that:

https://github.com/datenanfragen/website/blob/e9c5d122cf5b20678e387ec359cbbffd82da97a1/layouts/partials/summary.html#L2-L5

The fix should also be similar to the shortcode I proposed. Again, let me know if you want to do that or if I should.

> I was wondering, what does the `act` folder stands for? And should it be translated?

Those are our 'act pages' (I know the name isn't great but naming things is hard :D). They are essentially just regular blog posts that contain a call to some action. What makes them different is the special 'mini generator' that is included. They allow us to run 'mini campaigns' and respond very quickly to current events.  
For example, if there is a data leak, we can quickly spin up an act page allowing users to request their data from the company and ask if they are affected.

If a company like Facebook is covered by an act page, it would definitely make sense to translate it. Currently, we have only covered incidents in Germany, though, so I don't think there's much use in translating the ones we have right now.

> Moreover what about moving `verein` folder to `association` for better consistency with localization? Personally I think it can look weird for translators to have everything in English and a German word for this folder.

We definitely want to keep the `verein` in the German URLs. If you can find a way to translate that part of the URL however, I agree. That would be nice. Feel free to open an issue for that if you want.

> Finally, are you going to decline your material in different languages? If it's the case maybe it has to be moved to the content directory too?

Are you talking about https://www.datenanfragen.de/verein/material? That content is all hosted on our 'CDN' and not in this repo.

Regardless, we would love to have that available in multiple languages. However, that is not something Lorenz and me will be able to carry out, unfortunately. We are already struggling to provide material in German. But if someone wants to maintain our informational material in a different language, we'll be glad to help out as much as we can.

baltpeter:
> @baltpeter sure you can

Actually, it turns out it isn't quite as easy to force-push to a branch in a fork. :D I have opened #292 to continue this, now with a clean history.

You should still be able to push to that branch as you are a member of the repo.

NathanBnm:
> Thanks for changing the URLs manually. I think we should implement this shortcode before merging. If you're up for that, go for it, otherwise I can do that.

I'm not very familiar with Hugo so it's up to you, you can work on it, I mean I think you know how to proceed.

> Ok. Let's move the downloads and setup redirects for the old URLs.

Okay so I can work on this on the new PR, just don't know how to create redirections yet.

> I don't think it's feasible to try and contact all linking sites (for starters we don't have a list of them). I don't mind keeping the redirects around anyway, though, so that's fine.

Sure I understand.

> As long as we also setup redirects for the remaining ones, that's fine with me. I remember using these URLs in other places before (like [here](https://github.com/datenanfragen/material-hlci-2020/blob/master/kurse/brand-identity.md)), so they need to keep working.

You're right!

> Good catch! That's pretty much the same issue as with the images in the blog posts. The featured image URLs are now relative instead of absolute and the summary template doesn't know how to deal with that:
> 
> https://github.com/datenanfragen/website/blob/e9c5d122cf5b20678e387ec359cbbffd82da97a1/layouts/partials/summary.html#L2-L5
> 
> The fix should also be similar to the shortcode I proposed. Again, let me know if you want to do that or if I should.

Same here as I said I'm not very familiar with Hugo 😊

> Those are our 'act pages' (I know the name isn't great but naming things is hard :D). They are essentially just regular blog posts that contain a call to some action. What makes them different is the special 'mini generator' that is included. They allow us to run 'mini campaigns' and respond very quickly to current events.
>
> For example, if there is a data leak, we can quickly spin up an act page allowing users to request their data from the company and ask if they are affected.
> 
> If a company like Facebook is covered by an act page, it would definitely make sense to translate it. Currently, we have only covered incidents in Germany, though, so I don't think there's much use in translating the ones we have right now.

Understand now! And I agree for translation.

> We definitely want to keep the `verein` in the German URLs. If you can find a way to translate that part of the URL however, I agree. That would be nice. Feel free to open an issue for that if you want.

Sure! I think I already talked about it before. I'll try that once this will be finished.

> Regardless, we would love to have that available in multiple languages. However, that is not something Lorenz and me will be able to carry out, unfortunately. We are already struggling to provide material in German. But if someone wants to maintain our informational material in a different language, we'll be glad to help out as much as we can.

For sure! I can handle this for French, would be pleased to contribute 👍

PR #271, opened 2020-03-07
Fixes #256: Test for wizard regressions

### LINKED ISSUE #256: Implement test for #253
Apparently, our wizard test didn't catch #253. So, we need a better test.

baltpeter:


cypress/integration/use-cases/wizard.spec.js
```diff
@@ -57,5 +57,28 @@ describe('Using the wizard', () => {
...
+        cy.get('.request-transport-medium-chooser')
+            .contains('Fax')
+            .click();
+
+        cy.contains('Download PDF')
+            .should('not.have.class', 'disabled')
+            .clickLinkWithoutFollowingHref();
+
+        cy.contains('Next request').click();
+        cy.get('aside.company-info.box').get('.company-info-content');
```
zner0L: Why do you do that here? The way I understand it, this should do nothing, right?

baltpeter: Quite the opposite. If the company info box isn't there (which was one symptom of the bug), the test will fail. Almost all Cypress commands have implicit assertions built-in, makes it really easy to write tests.

zner0L: I wasn't aware of that. Good to know!

PR #281, opened 2020-03-31
use JSON.stringify()'s replacer to fix #187

### LINKED ISSUE #187: Match format of generated JSON on suggestion page and company-json-generator
This is a nice-to-have proposal. Feel free to put it deep down the backlog.

When a user adds a suggestion on https://www.datenanfragen.de/suggest, the generated JSON differs from the generated JSON of https://github.com/datenanfragen/company-json-generator.
Specifically, on the former, the slug is at the end, while on the latter it's at the beginning.

I would suggest to move the slug of the generated JSON of https://www.datenanfragen.de/suggest to the beginning, as order of the company-json-generator form items is mostly aligned with the generated JSON.

mal-tee:
My attempt at fixing #187 by using JSON.stringify()'s second parameter _replacer_.

### DIFF
--- a/src/suggest-edit.js
+++ b/src/suggest-edit.js
@@ -148,17 +148,17 @@ document.getElementById('submit-suggest-form').onclick = () => {
 
     document.getElementById('loading-indicator').classList.remove('hidden');
 
-    function getAllPropertyNamesinSchema(schema_part){
+    function getAllPropertyNamesInSchema(schema_part) {
         let res = [];
         for (let key in schema_part) {
-            if (key === "items"){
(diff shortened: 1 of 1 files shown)


src/suggest-edit.js
```diff
@@ -145,14 +147,42 @@ document.getElementById('submit-suggest-form').onclick = () => {
     if (data.fax) data.fax = formatPhoneNumber(data.fax);
 
     document.getElementById('loading-indicator').classList.remove('hidden');
+
+    function getAllPropertyNamesinSchema(schema_part){
```
baltpeter: The `i` in `in` should be capitalized.

mal-tee:
Great :)

PR #292, opened 2020-04-12
Restructure content directory (originally #269)

baltpeter:
This is a continuation of #269 to clean up the history.


baltpeter:
The quotes are getting a little unwieldy, so I'll just respond to [your comment](https://github.com/datenanfragen/website/pull/269#issuecomment-612623046) like this, @NathanBnm:

* Sure, I'll work on the Hugo image stuff.
* The redirects are setup in the aptly named [`_redirects`](https://github.com/datenanfragen/website/blob/master/_redirects). You can find the reference [here](https://docs.netlify.com/routing/redirects) or just [take a look at what I did for verein-website](https://github.com/datenanfragen/verein-website/blob/master/_redirects), that should pretty much cover it by example.
* As for the information material: Awesome! Currently, they are just InDesign files in a private repository. Not sure if you have access to InDesign but I'm sure we can also work out some other way. We should probably also move this to a new issue.

baltpeter:
Turns out, we need to do the restructure a little differently. Hugo has **very** silly requirements about this. With the current model, Hugo doesn't 'see' the image resources at all.

We actually need to implement the following structure (see [here](https://gohugo.io/content-management/page-bundles/)):

```
content
    en
        blog
            a-quick-overview.md
            sample-letter-whatever
                index.md
                some-image.jpg
                another-image.png
            supervisory-authorities
                index.md
                send-complaint.png
            your-gdpr-rights.md
```

I'll work on this.

NathanBnm:
You're right 😂

I'll take a look to redirections when I will have time. You can do so If I'm taking too long.

I can manage to have access to InDesign CC 2018. I think there are not so much difference between Adobe CC file versions.

I understand how the structure has to be. When a blog post has some ressources then it has to be turned into a folder. Am I right?

baltpeter:
> I'll take a look to redirections when I will have time. You can do so If I'm taking too long.

Great! No worries, as this will be a separate PR.

> I can manage to have access to InDesign CC 2018. I think there are not so much difference between Adobe CC file versions.

CC 2018 should definitely work.

> I understand how the structure has to be. When a blog post has some ressources then it has to be turned into a folder. Am I right?

Exactly.

---

A while ago we were talking about images that are identical between languages. I have found out that page resources are actually *shared* between different language versions of a post.

So, we can make it our convention to always store those images in the English version of the post and then still access them from all the other languages. One thing to note though: In order for this to work, **all** language versions need to be in a folder, even those without images. Otherwise, Hugo will not recognize that they are linked.

Not the cleanest solution, but better than duplicating the images.

baltpeter:
Phew, that wasn't quite as easy as expected. :D But at least now we have a solution that is quite clean from an end-user's perspective.

baltpeter:
Alright, I have cleaned up the commits. I think this should be ready for merge now once reviewed.

PR #296, opened 2020-04-15
Yearly report 2019

baltpeter:

PR #304, opened 2020-05-08
Finish updates to the transparency page

baltpeter:


baltpeter:
@NathanBnm If I remember correctly, you wanted me to mention you once we publish our yearly report for 2019. Took a little longer than we would've wanted but between #296 and this PR, the report has now been published and the transparency details updated. Would be great if you could update the French transparency page as well—luckily, it's only a couple of changes. :)

PR #308, opened 2020-05-09
Fixes #289: Offer copying the text in the MailtoDropdown manually 

### LINKED ISSUE #289: MailtoDropdown: Offer copying the text manually
In addition to the regular `mailto` link and the webmailers, it should be possible to get a modal displaying the subject, recipient and body that the user can copy manually.

baltpeter:


baltpeter:
Fixed some problems with the tests. This can be squashed when merging.

src/Components/MailtoDropdown.js
```diff
@@ -24,6 +31,77 @@ export const MAILTO_HANDLERS = {
...
+                            <input
+                                type="text"
+                                id="mailto-dropdown-copymanually-subject"
+                                className="form-element"
+                                value={decodeURIComponent(d.subject)}
+                                onClick={onInputClick}
+                                readOnly
+                            />
+
+                            {d.email
```
zner0L, line 69: I actually think conditionally hiding this is confusing. Why not just leave it empty? When I tested I was cunfused whether it was missing. An empty input would hint to users that they need to select a recipient.

src/Components/MailtoDropdown.js
```diff
@@ -49,10 +127,13 @@ export default class MailtoDropdown extends preact.Component {
         };
         const handler_buttons = handlers.map(h => (
             <a
-                href={MAILTO_HANDLERS[h].link(data)}
+                href={MAILTO_HANDLERS[h].link && MAILTO_HANDLERS[h].link(data)}
```
zner0L, line 130:
```suggestion
                href={MAILTO_HANDLERS[h].link?(data)}
```
After rebasing that can be simplified.

src/Components/MailtoDropdown.js
baltpeter: Hm, interesting. Personally I think showing the empty input looks quite stupid.

But I mean, eh. It's not the end of the world. Sure, why note.

src/Components/MailtoDropdown.js
baltpeter: Actually, that's `href={MAILTO_HANDLERS[h].link.?(data)}` as per [1](https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining.html#calling-deeply-nested-functions). :)

PR #313, opened 2020-05-14
Fix text overflowing in mobile browsers

baltpeter:
Mobile browsers don't seem to really support hyphenation unfortunately, which caused overflowing text. To avoid that, we now at least break the words on those platforms.

To test this: Check the German contribute page in a mobile browser.

### DIFF
--- a/content/de/contribute.md
+++ b/content/de/contribute.md
@@ -12,7 +12,7 @@ Vielen Dank, dass Du Dich über die Möglichkeiten zum Mitmachen informierst!
 <article id="cdb" class="list-article icon-list-article">
     <div class="col25 article-featured-image"><img class="image" src="/card-icons/company.svg" alt="Wartung der Unternehmensdatenbank"></div>
     <div class="padded col75">
-        <h1>Wartung der Unternehmensdatenbank</h1>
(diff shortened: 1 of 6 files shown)


zner0L:
Can you explain the percy change? Looks kinda ugly, but I am not sure whether that is just a percy related issue.

baltpeter:
This seems to be a Chrome problem. The same ugly rendering also occurs in regular Chrome on desktop. Apparently Chrome matches on `@supports not (hyphens: auto)`. :(

Not really sure what to do here. The motivator for this PR was when @mal-tee notified me that 'Unternehmensdatenbank' on the German contribute page doesn't break on mobile, causing the text to overflow.

It seems like hyphenation really isn't supported well in browsers. If we want to solve this generally, afaik my proposal in this PR is the only solution. But it causes these *very* ugly breaks in Chrome and on mobile. Far from great.  
The other option would be to simply add some `&shy;`s to that word and hope it doesn't occur anywhere else. Not a fan of either but I think I'm leaning towards the latter option. :/

zner0L:
Ugh, I hate browsers. I think having a broken footer on basically every page is extremely ugly. So I'd prefer the latter option, too.

baltpeter:
Updated this PR accordingly. I had a quick check around the site and didn't spot any other instances of this happening.

PR #320, opened 2020-05-26
Optimize Netlify deploys

baltpeter:
Recently, I saw [this post](https://community.netlify.com/t/support-guide-making-the-most-of-netlifys-cdn-cache/127) on the Netlify forums, explaining common problems with large sites causing long deploy times. And if we look at [two](https://app.netlify.com/sites/datenanfragen/deploys/5ec2bb3119ca5700074682b7) [recent](https://app.netlify.com/sites/datenanfragen/deploys/5ec2bbb6ea97040006b95cee) deploys, we see this exact phenomenon on our site: More than 5000 changed files have to be uploaded with each deploy.

That is clearly not a result of the [actual changes](https://github.com/datenanfragen/website/commit/d0928da2ce45d16311fd94211576271b75424297) between these two commits. Luckily, Netlify allows us to download the build results (takes a while though and sometimes needs a couple of tries). If we throw those into a directory diff tool, the problem quickly becomes apparent:

image

As I [already noted](https://github.com/datenanfragen/website/blob/2130c1bdcdd3a4f994b2c0c0add3da93a73237d1/layouts/company/single.html#L105) when introducing the related companies feature, the related pages Hugo generates are not deterministic at all. They change with every build, causing every entry in the company database to have a different hash each time.

This PR replaces the Hugo implementation with our own one. Unfortunately, this means the full Hugo build takes a fair bit longer. On my machine, running `hugo -e production --minify` takes:

* without related pages: `2748 ms`
* with the previous implementation: `3713 ms`
* with a one-line intersect: `7822 ms`  
  (`{{ $related := (where .Site.Pages ".Params.categories" "intersect" .Params.categories) | intersect (where .Site.Pages ".Params.relevant-countries" "intersect" (.Param "relevant-countries")) | first 5 }}`)
* with the more complex intersect loop and scoring: `18043 ms`

However, these build times are not really relevant as we have disabled the related companies calculation for local devs builds anyway, so we won't notice them on our machines. I am also fairly confident that the extra 10 seconds will be worth it in CI envs as currently on Netlify the actual build takes about 7 minutes while deploying and processing takes 14 minutes.

---

Note: This PR doesn't address another, analogous problem: We are using absolute URLs in our builds. That means that despite the changes in this PR, deploy previews will still cause huge diffs, at least on the first build. Subsequent builds should be faster.  
I don't believe it is worth it to change that, though. Especially since this doesn't affect production builds.

### DIFF
--- a/layouts/company/single.html
+++ b/layouts/company/single.html
@@ -102,20 +102,33 @@
         {{ T "cdb-help-needed" (dict "editUrl" (print ("suggest" | absURL) "?type=edit&for=cdb&slug=" .Params.slug) "newUrl" (print ("suggest" | absURL) "?type=new&for=cdb")) | safeHTML }}
     </div>
 
-    <!-- TODO: This is currently a very naive approach, mostly since we don't exactly have any data to do this better. We will want to improve this in the future. Oh yeah, and this feature seems to be very 'non-deterministic', at least under our parameters. Not sure if that is a good or a bad thing. -->
+    <!-- TODO: This is currently a very naive approach, mostly since we don't exactly have any data to do this better. We will want to improve this in the future. -->
     {{ if not .Site.Params.devMode }}
-    {{ $related := (where (.Site.RegularPages.Related .) "Type" "company") | first 5 }}
-    {{ with $related }}
-    <div id="related-companies">
-        <h2>{{ T "cdb-related-companies" }}</h2>
-        <!-- TODO: Nicer styling -->
-        <ul>
-            {{ range . }}
-            <li><a href="{{ .Permalink }}">{{ .Param "name" }}</a></li>
+        {{ $related_pages := slice }}
+
+        {{ $num_categories := (cond (isset .Params "categories") .Params.categories slice) | len }}
+        {{ $num_countries := (.Param "relevant-countries") | len }}
+
+        {{ range (where .Site.Pages "Section" "company") }}
+            {{ $num_common_categories := intersect $.Params.categories .Params.categories | len }}
+            {{ $num_common_countries := intersect ($.Param "relevant-countries") (.Param "relevant-countries") | len }}
+            {{ if and (gt $num_common_categories 0) (ne .Permalink $.Permalink) }}
+                {{ $score := add (div (float $num_common_categories) $num_categories) (div $num_common_countries $num_countries) }}
+                {{ $related_pages = $related_pages | append (dict "page" . "score" $score) }}
             {{ end }}
-        </ul>
-    </div>
-    {{ end }}
+        {{ end }}
+        {{ $related_pages = sort $related_pages "score" "desc" }}
+        {{ with $related_pages }}
+            <div id="related-companies">
+                <h2>{{ T "cdb-related-companies" }}</h2>
+                <!-- TODO: Nicer styling -->
+                <ul>
+                    {{ range (. | first 5) }}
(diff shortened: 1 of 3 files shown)


baltpeter:
We may also want to [ask them to completely disable post processing](https://community.netlify.com/t/completely-disabling-post-processing-of-deploys/2678) our our site. I'm pretty sure we are already doing all of that in our build pipeline anyway.

layouts/company/single.html
```diff
@@ -102,20 +102,33 @@
...
+        {{ $related_pages := slice }}
+
+        {{ $num_categories := (cond (isset .Params "categories") .Params.categories slice) | len}}
+        {{ $num_countries := (.Param "relevant-countries") | len}}
+
+        {{ range .Site.Pages }}
+            {{ $num_common_categories := intersect $.Params.categories .Params.categories | len }}
+            {{ $num_common_countries := intersect ($.Param "relevant-countries") (.Param "relevant-countries") | len }}
+            {{ if and (gt $num_common_categories 0) (gt $num_common_countries 0) (ne .Permalink $.Permalink) }}
+                {{ $score := div (add (div (float $num_common_categories) $num_categories) (div $num_common_countries $num_countries)) 2 }}
```
zner0L, line 116: It seems to me categories should be prioritized over countries, since having common countries is not really that useful for the users, is it?

layouts/company/single.html
```diff
@@ -102,20 +102,33 @@
...
+                {{ $related_pages = $related_pages | append (dict "page" . "score" $score) }}
             {{ end }}
-        </ul>
-    </div>
-    {{ end }}
+        {{ end }}
+        {{ $related_pages = sort $related_pages "score" "desc" }}
```
zner0L, line 120: Also, couldn't this be done in the first loop to save time?

layouts/company/single.html
baltpeter: Well, I'm pretty sure that's what the previous algorithm did anyway. Feel free to propose a different scoring system.

layouts/company/single.html
baltpeter: Not really sure how this would work to be honest. Running the build with and without this line only causes time differences well within the margin of error though. We are talking ±200 ms, with the winner alternating between the two, so it doesn't seem to make any significant difference.

baltpeter: Interestingly enough, changing the if clause to `{{ if and (gt $num_common_categories 0) (ne .Permalink $.Permalink) }}` consistently decreases build times by about a second, though.

That would mean showing companies with no related countries if there are none with related countries. I actually think I like that better anyway. What do you think, @zner0L?

baltpeter: There is also no actual reason to normalize the scores to [0, 1]. Removing that doesn't cause a measurable change in build times, though. Your call.

baltpeter:
I've [asked](https://community.netlify.com/t/completely-disabling-post-processing-of-deploys/2678/33?u=benni) for post-processing to be disabled now.

There already seems to be quite an improvement with the second deploy in this PR 'only' taking 13 mins and only 31 vs. 6000 files to upload!

image

baltpeter:
Post-processing should now be disabled. Excited to see what that will mean for the deploy times.

layouts/company/single.html
zner0L: I think we should do that then. Normalization can stay in there.

PR #322, opened 2020-05-30
Implement various security headers

baltpeter:
We weren't actually doing well on that front at all. See the previous result from [Webbkoll]):

image

This PR implements various security headers and also makes use of SRI for CDN resources. It should be tested **thoroughly** as these changes (especially the CSP) can easily break currently working behavior.

The CSP implemented by this PR isn't great as it still allows inline scripts and styles. Getting rid of the inline styles shouldn't be too much of a problem (we just need to find all of them). The scripts are a little more of a problem. Thus, I think we should defer this to a future PR.


deploy.sh
```diff
@@ -60,6 +60,7 @@ then
 	hugo -e production --minify
 	echo "Copying files for Netlify…"
 	cp _redirects public/_redirects
+    cp _headers public/_headers
```
baltpeter, line 63: While this looks off here, this is actually "correct". And switched the tabs in the surrounding lines to spaces in a different PR. That's why I am also using spaces instead of tabs here.

zner0L: It looks like something broke here. Is the " missing intentionally?

baltpeter: You broke that. :P 

https://github.com/datenanfragen/website/commit/6b55968593e46cc1b65dd5a5cdacc215393fac13#diff-60254338249f657a0a83f98258a56bfe

Issue #326, opened 2020-06-15
Donation via bank transfer allows negative values


mal-tee:
If you copy-paste a negative number into the donate form at https://www.datarequests.org/donate/  you can still procceed with *bank transfer*. The other payment methods raise an error.

baltpeter:
Thanks for letting us know!

This is a fairly non-critical issue that seems like a great first issue, so I'll leave it for someone else to work on. We've also recently done something similar in  #301, so that can serve as a guide.

medeasyJsr:
hey @baltpeter, I would like to help with this issue

baltpeter:
Sure, thank you!

Let me know if you have any questions.

kumar80:
Hey @baltpeter , I am working on this issue wanted to know if there should be any flash messages if yes what would be it

baltpeter:
Yes, a flash message would be good, @kumar80. You can use the same one as in #301 (see [here](https://github.com/datenanfragen/website/pull/301/files#diff-17b99fa8e6c12a09b25fbadc543db4f9R296)).

PR #327, opened 2020-06-17
Fixed issue that allowed donation of negative values via bank transfer  #326 

kumar80:
Checking for negative or zero values when the payment method is bank transfer also added flash error message


kumar80:
@baltpeter hey I have updated the list, Thank you very much.

PR #341, opened 2020-07-24
 Added symbol for external links, fixes #287 

### LINKED ISSUE #287: Style external links differently
To make it more clear to users when a link leads to a website not controlled by us, we should style external links differently. This is commonly done using a trailing icon, like [here](https://www.kuketz-blog.de/datenschutzhinweis-kuketz-meet-de/):

image

To be clear, "internal" and "external" in this context are **not** to be understood technically like with *origins* on the web but rather as websites controlled vs. not controlled by us. This will be probably have to be implemented as a list anyway.

"Internal" websites I can think of (there may be more):

* *.datenanfragen.de
* *.datarequests.org
* *.demandetesdonnees.fr
* *.pedidodedados.pt
* *.dacdn.de

mal-tee:
Please rebase onto master after merging #334


baltpeter:
We missed one case: `#anchor` links. :D 

image

baltpeter:
Oh, and we should probably also include `--datenanfragen.netlify.app/`. This will unfortunately need to be matched with a __contains__ instead of a __starts with__ selector. :(

src/styles/framework.scss
```diff
@@ -200,32 +200,33 @@ span > a,
...
-            bottom: 0;
-            background-color: $accent-color;
-            @media (prefers-reduced-motion: no-preference) {
-                transition: all ease 0.3s;
+        // Very loosely inspired by: https://codepen.io/aaroniker/pen/pojaBvb
+        background-image: linear-gradient(0deg, $accent-color 0%, $accent-color 100%);
+        background-position: bottom left;
+        background-repeat: no-repeat;
+        background-size: 100% 2px;
+        $external-link-selector: '&:not([href^="https://www.datenanfragen.de"]):not([href^="https://www.datarequests.org"]):not([href^="https://www.demandetesdonnees.fr"]):not([href^="https://www.pedidodedados.org"]):not([href^="https://static.dacdn.de"]):not([href^="http://localhost:"]):not([href^="http://datenanfragen.test:"]):not([href^="https://verein.datenanfragen.de"]):not([href^="#"]):not([href*="--datenanfragen.netlify.app/"])';
```
zner0L, line 212: Wouldn't `variables.scss` be a better place for that?

zner0L: Why include `localhost` and why only http? For testing? When the changes are live, localhost would be external IMO.

src/styles/framework.scss
```diff
@@ -200,32 +200,33 @@ span > a,
...
 
         &:hover,
         &:focus {
             color: color('white');
-
-            &::after {
-                height: 100%;
+            background-size: 100% 100%;
+            #{$external-link-selector} {
+                background-size: calc(100% - 16.8px) 100%;
```
zner0L, line 229: Why exclude the symbol from the background?

mal-tee: This way it is not affected by the hover animation. Otherwise the animation would cover the symbol as well, which would defeat its purpose, if we do not change its color, and does not look good imo.

src/styles/framework.scss
baltpeter: > Wouldn't variables.scss be a better place for that?

Imho this is the better place. `variables.scss` is for 'global' variables that are used in multiple places. This doesn't apply here. `$external-link-selector` is only used here and I don't see a case where it would be needed somewhere else.

baltpeter: >  Why include localhost and why only http? For testing? When the changes are live, localhost would be external IMO.

The only instance of `localhost` links appearing on the site is when developing locally. In this case, these are indeed internal links. We have always treated `localhost` as such, it also has CORS access for example.

src/styles/framework.scss
baltpeter: I agree with @mal-tee that including the symbol in the underline looks worse. See also the reference provided in #287.

src/styles/framework.scss
mal-tee: * regarding `http`: Because the dev-server unfortunately doesn't work with `https`

zner0L: > Imho this is the better place. `variables.scss` is for 'global' variables that are used in multiple places. This doesn't apply here.

While I agree the selector isn't global, I disagree with the purpose of `variables.scss`. IMO, it is the place for global varibales or all preference-ish variables, so that if you want to tweak the styles to another website, everything you need to change is in there. Context-wise that makes more sense to me. But I don't care much about that.

baltpeter: I don't care too much either, so since @mal-tee already changed it, we might as well leave it. *shrug*

zner0L: I feel like, not marking localhost as an external address in production might be a problem in spotting wrong links that appeared correct during local testing. Also, an attacker could use this to place links to localhost on our website that would then appear to be internal. But I guess in that case that would be the least problem. Still, it doesn't feel right to me.

zner0L: I'll let @mal-tee decide, whether they want to change it or not.

baltpeter: > I feel like, not marking localhost as an external address in production might be a problem in spotting wrong links that appeared correct during local testing.

I don't see this as a diagnostic tool for use but as an indicator for users on whether a link will lead them to a different website. If you want a tool like that, I wouldn't be opposed to that but wouldn't it be easier to just grep `public` for `localhost` or something similar? That seems way easier and more effective.

 > Also, an attacker could use this to place links to localhost on our website that would then appear to be internal.

I don't really see the attack vector, to be honest. How would they place this link? In that case they would most likely have XSS already anyway, wouldn't they?

zner0L: > I don't really see the attack vector, to be honest. How would they place this link? In that case they would most likely have XSS already anyway, wouldn't they?

Yes, but a link to localhost would maybe allow them to escalate their attack. I just see no reason to leave this avoidable pattern in, other than comfort. It just seems a little sloppy to have dev-only code in production. I know that there is no real risk here and I am fine with merging anyway.

mal-tee: I agree with Benjamin, but I can see why you do not want `localhost` in the list. How about we merge this as it is and create a new Issue. Then @zner0L or somebody else can implement it later on.

mal-tee: Btw: If it was possible to inject a link, an attacker could inject it in such a way that the selector for this symbol is not applicable.

mal-tee: I.e. they could be able to circumvent this indicator even without `localhost` in this list.

PR #344, opened 2020-08-03
Fixes #342: Switch to the new Typesense 0.14 instance

### LINKED ISSUE #342: Search only works for `runs` list in some cases if country is set to `all`
Company: https://www.datarequests.org/company/match-group

Peek-2020-07-28-14-33

baltpeter:
Before this can be merged, the following has to happen:

* datenanfragen/data#651 needs to be merged
* `search.datenanfragen.de` needs to point to the new instance
* this PR needs to be updated to not use `search-beta.datenanfragen.de` anymore (and the `WIP` needs to be removed from the commit message)

### DIFF
--- a/src/Components/SearchBar.js
+++ b/src/Components/SearchBar.js
@@ -20,7 +20,7 @@ if (Privacy.isAllowed(PRIVACY_ACTIONS.SEARCH)) {
                     host: 'search.datenanfragen.de',
                     port: '443',
                     protocol: 'https',
-                    apiKey: '981da30f2e80463cadf8a91bc2db4cf5'
+                    apiKey: ''
                 },
                 timeoutSeconds: 2
             });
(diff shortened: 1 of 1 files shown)


src/Components/SearchBar.js
```diff
@@ -45,7 +45,7 @@ if (Privacy.isAllowed(PRIVACY_ACTIONS.SEARCH)) {
         componentDidMount() {
             let options = {
                 query_by: this.props.query_by || 'name, runs, categories, web, slug, address, comments',
-                sort_by: 'sort-index:asc',
+                sort_by: '_text_match:desc,sort-index:asc',
```
zner0L, line 48: I can't find `_text_match` in the documentation. Did you mean `__match_score`?

baltpeter: Interesting. The [guide](https://typesense.org/docs/0.14.0/guide/) talks about `_text_match` while [ŧhe API docs](https://typesense.org/docs/0.14.0/api/) talk about `__match_score`. Not sure what the difference is. `_text_match` seems to work fine and do what we want, though.

zner0L: Weird. The guide apparently mentions both, `_text_match` and `__match_score`. But it doesn't differentiate them. That worries me a bit, but ok. If it works, it should be fine I guess…

zner0L: Well, apparently using `_ match_score` produces an error, so lets just keep it that way.

PR #349, opened 2020-08-07
Add pedidodedados.org to the verein page

mal-tee:


src/i18n/README.md
```diff
@@ -13,3 +13,4 @@
 * Translate the 404 page.
 * Setup the necessary redirects in `_redirects`.
 * Add the domain in the Netlify control panel and set the appropriate DNS records.
+* Add the domain to `content/de/verein/_index.md`
```
baltpeter, line 16: Please move this line up a little. While the steps aren't really _ordered_ per se, it feels wrong to have this as the last step.

mal-tee:
I can't speak French, so I would prefer if someone else could change the French page. 😅

PR #351, opened 2020-08-07
Load the pdfworker.js only if it is needed, fixes #330

### LINKED ISSUE #330: Make the generator lighter by loading heavy parts on demand
The PDF-part of the generator makes up a significant part of the generator-script, but is only used for non-email requests. I propose that we load the PDF-code on demand, e.g. if the user switches to another mode than email or interacts with the generator in another way.
This will reduce the size of the core-script and therefore improve loading times and performance.

mal-tee:
* I just moved the initialization of the pdfworker into the `initiatePdfGeneration`
* If the state changes to fax or letter the `pdfworker.js` is loaded
* Otherwise it is not loaded anymore, saving us some traffic and time

### DIFF
--- a/cypress/integration/components/generator.spec.js
+++ b/cypress/integration/components/generator.spec.js
@@ -1,0 +1,16 @@
+describe('Generator component', () => {
+    beforeEach(() => {
+        cy.visit('/generator');
+    });
+
+    it('has generated a blob URL', () => {
+        cy.get('.request-transport-medium-chooser')
+            .contains('Fax')
+            .click();
+
+        cy.contains('Download PDF')
+            .should('not.have.class', 'disabled')
+            .should('have.attr', 'href')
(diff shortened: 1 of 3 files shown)


cypress/integration/components/generator.spec.js
```diff
@@ -0,0 +1,39 @@
...
+    it('did load pdfworker for fax', () => {
+        cy.get('.request-transport-medium-chooser')
+            .contains('Fax')
+            .click();
+
+        cy.window().then(win => {
+            expect(win.pdfWorker).not.to.be.undefined;
+            // somehow it does not work with this: expect(win.pdfWorker).to.satisfy((o) => o instanceof Worker);
+            // "console.log(win.pdfWorker instanceof Worker)" logs false...
+            // time to 🦆 type...
```
zner0L, line 32: How do you like emojis in our comments, @baltpeter? 😜

baltpeter: Big fan!

PR #353, opened 2020-08-08
Fixes #352: Properly translate categories in the search

### LINKED ISSUE #352: Search falsely returns empty categories for "schokol"
grafik
grafik

* Also works for "schokola" 🍫 
* Is not fixed by #344, seems to be a parsing error on our side

baltpeter:
This commit doesn't solve the issue per se but rather removes the mechanism that caused it as there was really no reason for that.

Let me explain: Previously, the search also queried the categories of the records. Therefore, `schokol` matched the category "school" (considering an allowed Levenshtein distance of 4).

For the matches, Typesense helpfully returns a "highlights" property where our category now looks like this: `<mark>school</mark>`.

The problem was that we were also feeding these "highlighted" results into our translate function which obviously didn't return anything as we don't have a translation for `<mark>school</mark>`.

This left me with two options: Either implement some annoying way to translate while still preserving the highlights or just drop searching in categories.*

The choice was quite obvious given that I don't really see any use for searching in the categories, **especially** the untranslated ones.

---

\*Technically, just removing `/<\/?mark>/` for the translation and then re-appending it around the whole translated word would have _probably_ been a good enough approximation and therefore another option.

  However, given that it really doesn't make sense to be able to search the categories anyway, this doesn't matter.

### DIFF
--- a/src/Components/SearchBar.js
+++ b/src/Components/SearchBar.js
@@ -44,7 +44,7 @@ if (Privacy.isAllowed(PRIVACY_ACTIONS.SEARCH)) {
 
         componentDidMount() {
             let options = {
-                query_by: this.props.query_by || 'name, runs, categories, web, slug, address, comments',
+                query_by: this.props.query_by || 'name, runs, web, slug, address, comments',
                 sort_by: 'sort-index:asc',
                 num_typos: 4,
                 per_page: this.props.numberOfHits || 5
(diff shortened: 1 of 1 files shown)


baltpeter:
As a result, `schokol` doesn't return anything anymore but I feel like that is a good thing since the previous results weren't exactly… helpful given the query. :D 

image

zner0L:
Can you explain the percy change, @baltpeter?

baltpeter:
> Can you explain the percy change, @baltpeter?

The translation for the empty list in the wizard changed. The 'Why don't you add some?' was added after this PR was submitted.

PR #356, opened 2020-08-09
Switch to the new backend

baltpeter:


baltpeter:
Actually, I have changed my mind about the draft status. We should get this merged asap, there are already new comments coming in an the longer we wait the more annoying this will get. We can switch over the donate backend in a separate PR once that is ready.

Issue #360, opened 2020-08-13
External links cleanup


baltpeter:
In #341, we added different styling for external links. Some artifacts of this that need to be cleaned up:

* Pure anchors without an href (as seen [here](https://www.datarequests.org/supervisory-authorities/)):

![image](https://user-images.githubusercontent.com/4048809/90184078-17174f00-dda4-11ea-80de-5772a685504e.png)

* mailto URLs (as seen [here](https://www.datarequests.org/contact)):

![image](https://user-images.githubusercontent.com/4048809/90184204-4463fd00-dda4-11ea-8044-23f2fe625c08.png)

* relative URLs (as seen [here](https://www.datarequests.org/contact)):

![image](https://user-images.githubusercontent.com/4048809/90184318-71181480-dda4-11ea-9569-c1e68810769e.png)

* whatever the hell is happening [here](https://www.datarequests.org/privacy#controller-and-contact-information):

![image](https://user-images.githubusercontent.com/4048809/90184389-91e06a00-dda4-11ea-843a-9ba1f13e10e1.png)

baltpeter:
* Another case, reported by @brnt-toast in #364: Manually added anchors (as seen [here](https://www.datarequests.org/supervisory-authorities)):  
  
  
![image](https://user-images.githubusercontent.com/4048809/90569120-cb4a1880-e19c-11ea-85db-b5fccd42d357.png)

PR #364, opened 2020-08-18
deleted two dead links

brnt-toast:


zner0L (commented):
Those are anchors that are needed in order to link to this specific part of the article using URL fragments. Take a look at the [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Linking_to_an_element_on_the_same_page) if you want to learn more about this.

baltpeter:
Hi @brnt-toast, thanks for the PR! As @zner0L said, we do in fact need those anchors. But you are of course correct that they shouldn't show up on the page.

This is caused by the separate styling for external links we recently added (#341). I have already opened an issue with some of the problems caused by that (#360) but we have missed this case. I'll add it to the list. :)

PR #368, opened 2020-08-23
Production tests

baltpeter:


baltpeter:
This is also ready.

cypress/integration/components/generator.spec.js
```diff
@@ -1,31 +1,33 @@
...
+        cy.window().then((win) => {
             expect(win.pdfWorker).to.be.undefined;
         });
     });
 
     it('did load pdfworker for fax', () => {
-        cy.get('.request-transport-medium-chooser')
-            .contains('Fax')
-            .click();
+        skipOn(isOn('production'));
```
zner0L, line 26: Why is this test skipped?

cypress/integration/components/generator.spec.js
```diff
@@ -1,31 +1,33 @@
...
+        // close.
+        cy.contains('Download PDF', { timeout: 10000 })
             .should('not.have.class', 'disabled')
             .should('have.attr', 'href')
             .and('match', /^blob:https?:\/\/[\S]+?\/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/);
     });
 
     it('did not load pdfworker for email', () => {
-        cy.window().then(win => {
+        cy.window().then((win) => {
```
zner0L, line 20: Do we have access to `cy.window()` in production?

cypress/integration/use-cases/production.spec.js
```diff
@@ -0,0 +1,35 @@
...
+            cy.visit(`${site}/generator`);
+            cy.get('[for="request-type-choice-erasure"]').click();
+            cy.get('[for="request-transport-medium-choice-fax"]').click();
+            cy.get('#1-delete-id_data').click();
+            cy.get('#download-button').clickLinkWithoutFollowingHref();
+
+            cy.visit(`${site}/blog`);
+            cy.get(':nth-child(1) > .padded > a > h1').click();
+            // This assumes that we always have at least one blog post per language which might not be true in the
+            // future. We would then need to exclude those cases here.
```
zner0L, line 31: Maybe add this comment to the i18n documentation?

zner0L (commented):
If you can clear up the questions, this can be merged.

cypress/integration/components/generator.spec.js
baltpeter: Yes. Why wouldn't we?

cypress/integration/components/generator.spec.js
baltpeter: Because `window.pdfWorker` is only set in dev.

cypress/integration/components/generator.spec.js
zner0L: I guess this should be skipped in production, according to https://github.com/datenanfragen/website/pull/368/files#r484395914

baltpeter: Since it's never set, it will always be `undefined` and the test will always pass. But sure, whatever.

zner0L (approved):

PR #375, opened 2020-08-25
Use hugo pipes to generate CSS assets

zner0L:
This is still very much work in progress. There seems to be [a bug in the `PostCSS` pipe](https://discourse.gohugo.io/t/problem-chaining-tocss-and-postcss/24260/6) which might make this PR impossible. We also might loose *some*, however very limited, comfort features and gain a lot of build speed. But all of that is still very experimental atm.


zner0L:
I am still working on moving the js build to hugo as well, but for now, moving the scss is self-contained and can be merged. Please be aware that building now requires the **extended** version of hugo.

baltpeter (changes requested):
* The permalinks Hugo generates are always from the German site which means that they cannot be loaded on the other language versions due to CORS.
* There seems to be an option to include source maps. ([1](https://gohugo.io/hugo-pipes/scss-sass/#options)). Can we enable that please?

zner0L:
I fixed the permalinks, @baltpeter.

layouts/_default/baseof.html
```diff
@@ -31,9 +31,10 @@
     {{ if eq .Type "generator" -}}
     <!-- TODO: Implement this correctly! <link rel="preload" as="script" href="{{ "js/pdfworker.gen.js" | absURL }}"> -->
     {{- end }}
-    <link rel="stylesheet" href="https://static.dacdn.de/fonts/fira-sans.css" integrity="sha384-tdesaAgh/bq3iJZEqjfYp6o6sVopRLnw3wjZKW9gLPpNd26Pdgbur5PChCva/Z+2" crossorigin="anonymous">
-    <link rel="stylesheet" href="https://static.dacdn.de/fonts/league-spartan.css" integrity="sha384-n561KFrMi6x/rwAxQhsRLmxmF+oKREXWnteNZCFbPbU1nODnUri2U+r9p/lCJ918" crossorigin="anonymous">
-    <link rel="stylesheet" href="{{ "css/style.gen.css" | absURL }}" />
+    <link rel="stylesheet" href="https://static.dacdn.de/fonts/fira-sans.css" integrity="sha384-U8TLqRd5TXsbHc89iPjGqbk/4KrbgxDjzI37GGTuQ30EdFX0pgdZZzPJPLkL2iJR" crossorigin="anonymous">
+    <link rel="stylesheet" href="https://static.dacdn.de/fonts/league-spartan.css" integrity="sha384-8xT2dJ0AqzYTGWPC27D6SB6ndu+drRpQGL4BAxDYQvC87Adakg1LrWQY5t84tkkO" crossorigin="anonymous">
```
baltpeter, line 35: The rebase failed here. THE SRIs have changed:

https://github.com/datenanfragen/website/blob/9195f52c9366844e6fe8c08fe88952776eb29c05/layouts/_default/baseof.html#L34-L35

baltpeter (changes requested):

postcss.config.js
```diff
@@ -0,0 +1,31 @@
+const path = require('path');
+
+module.exports = (ctx) => {
+    return {
+        plugins: [
+            require('postcss-fonticons')({
+                iconPath: path.resolve(__dirname, 'assets/icons/'),
+            }),
+            require('postcss-preset-env')({
+                browsers: ['last 2 versions', '> 5%'],
```
baltpeter, line 10: This is duplicated from the `package.json`:

https://github.com/datenanfragen/website/blob/9195f52c9366844e6fe8c08fe88952776eb29c05/package.json#L114-L117

Since this is a JS file, can't we just import that here? Or, if the p.json entry isn't needed anymore, we should remove that.

But this is not a blocking problem, I'll create an issue.

baltpeter (approved):
Awesome, thanks. Very excited to see us finally working on a better build process.

postcss.config.js
haardikdharma10: @baltpeter, should we consider removing it?

PR #378, opened 2020-08-25
🔗 [Validation]: Unclosed div tag with id #home-features has a closing tag now.

rohitjethoe:
HTML Validation. 

Closing `</div>` tag added. It was missing before.


baltpeter (approved):
Thanks!

Issue #207, opened 2019-11-28
Footer is missing on 404 page


NathanBnm:
I think it would be nice to keep it for more consistency and to guide the user if he is lost on the site.

baltpeter:
I absolutely agree that it would be much better to have a footer on the 404 page. Unfortunately, not having it there wasn't a decision on our part rather than a necessary compromise to have a customized 404 page at all.

Basically, the problem is that we couldn't find a way to have Hugo generate the 404 page. Instead we can only have a single HTML file for all languages. See [this issue](https://git.my-server.in/datenanfragen/website/issues/115) in our old issue tracker for details.

baltpeter:
I'll leave this issue open in case anyone has an idea on how to solve this problem.

baltpeter:
Good news: We will soon be switching from Netlify to our own static site hosting solution (dattel, see #363). We should then be able to implement this.

PR #381, opened 2020-08-25
🇺🇳 [Validation]: Added lang attribute to <html> tag for all languages.

rohitjethoe:
### Validated html

Added a html lang attribute to the html tag in the right page language with JavaScript.

### DIFF
--- a/static/404.html
+++ b/static/404.html
@@ -81,6 +81,26 @@ if(window.location.host.includes('datenanfragen.de') || window.location.host.inc
 if(window.location.host.includes('demandetesdonnees.fr') || window.location.host.includes('1315')) lang = 'fr';
 if(window.location.host.includes('pedidodedados.org') || window.location.host.includes('1316')) lang = 'pt';
 
+// Adds 'lang' attribute to <html>
+let htmlSelector = document.querySelector('html');
+switch(lang) {
+    case 'en':
+        htmlSelector.setAttribute('lang', 'en-us');
+        break;
+    case 'de': 
+        htmlSelector.setAttribute('lang', 'de-de');
+        break;
+    case 'fr':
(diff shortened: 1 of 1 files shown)


baltpeter:
Hi,

thanks for the suggestion!

We will soon move to a new hosting solution that will allow us to have proper per-language 404 pages that don't need JavaScript for the translation (see #207). We will then be able to just add the `lang` attribute to the `<html>` tag directly.

Do you think this can wait until then? Otherwise, I would request some changes and propose a cleaner implementation. I'll leave that up to you. :)

rohitjethoe:
@baltpeter There isn't any other JavaScript file linked to the 404 page so I think this is currently the cleanest implementation.

static/404.html
```diff
@@ -81,6 +81,26 @@ <h1 class="translatable">
 if(window.location.host.includes('demandetesdonnees.fr') || window.location.host.includes('1315')) lang = 'fr';
 if(window.location.host.includes('pedidodedados.org') || window.location.host.includes('1316')) lang = 'pt';
 
+// Adds 'lang' attribute to <html>
+let htmlSelector = document.querySelector('html');
```
baltpeter, line 85: The region subtag is optional in the language tag (and it's not like we are targeting any specific regional language variant). Thus, we can do without the switch case. That's what I meant by a cleaner implementation. :) In fact, we can simplify the whole thing to one line:

```js
document.documentElement.lang = lang;
```

baltpeter (changes requested):

baltpeter:
As I mentioned, now that we have moved to another hosting solution, we don't need the hacky 404 implementation anymore. This PR is thus superseded by #398.

PR #383, opened 2020-08-26
Add the transcript for the OKL talk

baltpeter:
I had to write that anyway for the subtitles, so we might as well publish it for accessibility and SEO reasons.


zner0L (approved):
I sure use "so" a lot when talking… Thanks for the transcript!

PR #384, opened 2020-08-26
🖼 [Validation]: Logo/duplicate ids

rohitjethoe:
Fixed logo issue with duplicate id's.

This can be unnecessary with the new website incoming but for now this can be a good solution?

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -46,7 +46,7 @@
         <div id="nav-header-bg"></div>
         <div id="header-main" class="container">
             <a href="{{ "/" | absURL }}" title="{{ T "logo-a-title" . }}" style="position: relative;">
-                <img id="logo" src="{{ T "logo" . | absURL }}" title="{{ T "logo-title" . }}">
+                <img class="logo" src="{{ T "logo" . | absURL }}" title="{{ T "logo-title" . }}">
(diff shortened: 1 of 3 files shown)


layouts/_default/baseof.html
```diff
@@ -46,7 +46,7 @@
         <div id="nav-header-bg"></div>
         <div id="header-main" class="container">
             <a href="{{ "/" | absURL }}" title="{{ T "logo-a-title" . }}" style="position: relative;">
-                <img id="logo" src="{{ T "logo" . | absURL }}" title="{{ T "logo-title" . }}">
+                <img class="logo" src="{{ T "logo" . | absURL }}" title="{{ T "logo-title" . }}">
```
baltpeter, line 49: Thanks for the PR. There is really only supposed to be one logo per page. The 404 page is a temporary exception that will soon not be relevant anymore. Thus, I would rather not change this from an `id` to a `class` as it makes less sense semantically.

If you think that it is necessary that we fix this now, please just allow both cases in the CSS and keep this file as-is.

baltpeter (changes requested):

baltpeter:
As I mentioned, now that we have moved to another hosting solution, we don't need the hacky 404 implementation anymore. This PR is thus superseded by #398.

PR #385, opened 2020-08-27
🔗 [Validation]: Added alt attribute to the image of a blog post.

rohitjethoe:
Added alt attribute with the title as content in the attribute.


baltpeter (approved):
Good catch, I missed that in #372. Thanks!

PR #390, opened 2020-09-01
Fixes #388: Only load saved signature if there is one

### LINKED ISSUE #388: Generator doesn't work if "Remember identification data" is disabled: t.signature is null
A user reported that they used the wizard and were greeted with the below error upon clicking "Done adding companies" and being redirected to the generator.

```js
{
"type": "unhandledrejection",
"error": {
"message": "t.signature is null",
"name": "TypeError",
"stack":
"value@https://www.datarequests.org/js/generator.gen.js:29:41101\nn/<@https://www.datarequests.org/js/generator.gen.js:29:75824\nn/</<@https://www.datarequests.org/js/generator.gen.js:29:70764\n"
},
"defaultPrevented": false,
"eventPhase": 2,
"isTrusted": true,
"returnValue": true,
"code_version": "1.0.0",
"user_agent": "<redacted>",
"url": {
"href":
"https://www.datarequests.org//generator#!from=wizard&companies=<redacted>",
"origin": "https://www.datarequests.org",
"protocol": "https:",
"host": "www.datarequests.org",
"hostname": "www.datarequests.org",
"port": "",
"pathname": "//generator",
"search": "",
"hash":
"#!from=wizard&companies=<redacted>"
}
}
```

baltpeter:


src/Components/RequestGeneratorBuilder.js
```diff
@@ -79,13 +79,13 @@ export default class RequestGeneratorBuilder extends preact.Component {
...
-                prev.request.signature = res.signature;
+        await initializeFields(this.state.request.id_data).then((res) => {
+            this.setState((prev) => {
+                if (res.new_fields) prev.request.id_data = res.new_fields;
+                if (res.signature) prev.request.signature = res.signature;
 
-                const name = res.new_fields.filter(i => i.type === 'name')[0];
-                const address = res.new_fields.filter(i => i.type === 'address')[0];
+                const name = res.new_fields.filter((i) => i.type === 'name')[0];
+                const address = res.new_fields.filter((i) => i.type === 'address')[0];
```
zner0L, line 88: Why not include those in the if as well? If there is no `res.new_fields` this leads to problems as well.

zner0L (changes requested):
Thanks for the quick fix! Can you add in the the small change? Then we are good to go.

src/Components/RequestGeneratorBuilder.js
baltpeter: I'm reasonably certain that `SavedIdData.mergeFields()` will always return an array but sure, we might as well.

zner0L (approved):

PR #392, opened 2020-09-08
Privacy policy: We have changed our hosting providers

baltpeter:
This is to be merged after #363.


zner0L (changes requested):
Thanks for looking through that! It is nice to see we are finally getting rid of logging. Before this is merged we need to get rid of Exoscale, though, (or mention them here).

zner0L (approved):
Thanks for changing all that!

PR #400, opened 2020-09-17
Various suggest improvements

baltpeter:
One review is enough, whoever comes first. :D But I would like to get this merged sooner rather than later because the suggest box currently looks pretty stupid due to the new fields that I added in the schema.


src/Utility/common.js
```diff
@@ -17,10 +16,10 @@ export function fakeEvt(value) {
 }
 
 // Adapted after: https://stackoverflow.com/a/8498629
-export function domainFromUrl(url) {
+export function domainWithoutTldFromUrl(url) {
     if (!url) return;
 
-    let matches = url.match(/^https?:\/\/([^#/:?]+)(?:[#/:?]|$)/i);
+    const matches = url.match(/^https?:\/\/([^#/:?]+)\.([^#./:?]+)(?:[#/:?]|$)/i);
```
mal-tee, line 22: We might be able to simplify this

mal-tee:
Great! Thank you!

mal-tee (approved):

Issue #373, opened 2020-08-24
Use modern image formats with Hugo


baltpeter:
Google PageSpeed (rightfully) complains that we are not making use of [modern image formats](https://web.dev/uses-webp-images/):

![image](https://user-images.githubusercontent.com/4048809/91093425-f1faca00-e648-11ea-9b19-88a9a3f8af98.png)

While we could manually convert and use them, that would be very time consuming and easy to forget, especially since Hugo already [supports converting images automatically](https://gohugo.io/content-management/image-processing/). Unfortunately though, [Hugo doesn't support any modern image format like WebP yet](https://github.com/gohugoio/hugo/issues/5924). :(

If that is supported one day, we should absolutely [make use](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#Use_modern_image_formats_boldly) of such a feature.

Note that we can unfortunately not completely switch to WebP yet as [browser support](https://caniuse.com/webp) isn't sufficient. We still need to support the legacy formats using srcset, similar to what I did in #372.

rahulsunil2:
I've fixed this, [Pull Request](https://github.com/datenanfragen/website/pull/415)

baltpeter:
A quick note for everyone: As I said, we are **not** looking for manual conversions, only automatic ones as part of our Hugo build.

This issue cannot be resolved until support for that is added in Hugo.

chrysillala:
Hi, can I try to fix this issue? Thanks

baltpeter:
> Hi, can I try to fix this issue? Thanks

Please see my above comment: https://github.com/datenanfragen/website/issues/373#issuecomment-702612156

baltpeter:
Looks like we can finally tackle this soon: https://github.com/gohugoio/hugo/pull/8397

opncow:
From looking at the code of gohugoio/hugo#8397 I cannot tell: Will it be possible to convert e.g. JPEG images to webp automatically during the image processing? That would be neat!

baltpeter:
> Will it be possible to convert e.g. JPEG images to webp automatically during the image processing? That would be neat!

That's my understanding, yes (see also: https://github.com/gohugoio/hugo/issues/5924).

PR #415, opened 2020-10-01
Use modern image formats - .webp

rahulsunil2:
Issue: #373 

I have converted all the heavy modern image format files(.png, .jpeg) to .webp.
I've also made necessary changes in the code to replace the image src also.


baltpeter:
Thanks for taking the time to convert them manually. Unfortunately we explicitly don't want that done manually as I have said in #373: "While we could manually convert and use them, that would be very time consuming and easy to forget, especially since Hugo already supports converting images automatically."

Also, we can unfortunately not completely switch to WebP yet as [browser support](https://caniuse.com/webp) isn't sufficient. We still need to support the legacy format using srcset, similar to what I did in #372.

PR #420, opened 2020-10-02
Fixes #399: Add manifest-src=self to header of website

### LINKED ISSUE #399: Allow webmanifest in CSP
When loading the site in Chrome, the following error is logged to the console:

> Refused to load manifest from 'https://www.datenanfragen.de/site.webmanifest' because it violates the following Content Security Policy directive: "default-src 'none'". Note that 'manifest-src' was not explicitly set, so 'default-src' is used as a fallback.

Techievena:


Techievena:
@baltpeter Please review

baltpeter (approved):
Thanks!

PR #426, opened 2020-10-04
Applied link styling by modifying link selector.

### LINKED ISSUE #366: 'Fancy' link styling doesn't apply to all links
The link selector @zner0L implemented a while back isn't sufficient to catch all links.

@mal-tee mentioned two cases in #365:

* The links to GitHub in https://www.datarequests.org/privacy#error-reports
* The links in the author box, e.g. to creativecommons.org in https://www.datarequests.org/blog/gdpr-territorial-scope#page-meta

botex98:
Fixes #366 .

Screenshot from 2020-10-04 11-55-59

### DIFF
--- a/assets/styles/framework.scss
+++ b/assets/styles/framework.scss
@@ -194,10 +194,8 @@ a:focus {
     color: color('teal', 800);
 }
 
-p > a,
-span > a,
-.box a {
-    &:not(.button):not(.no-link-decoration) {
+main a {
+    &:not(.button):not(.no-link-decoration):not(.icon) {
         position: relative;
         text-decoration: none;
         @media (prefers-reduced-motion: no-preference) {
--- a/layouts/company/single.html
+++ b/layouts/company/single.html
@@ -9,7 +9,7 @@
 {{ define "main" }}
 <main class="narrow-page" style="overflow-wrap: break-word;">
     <div style="display: flex; align-items: center; justify-content: space-between;">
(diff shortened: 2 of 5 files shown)


baltpeter:
Thanks for the changes! Using `main` is a really clever idea. I'll leave the review to @zner0L but one case I noticed: The thumbnails in the [blog post list](https://deploy-preview-426--datenanfragen.netlify.app/blog) should not be styled. It should probably be safe to exclude all images, shouldn't it?

image

zner0L (changes requested):
Good work, thank you very much! As @baltpeter already pointed out, images need to be excluded, too. Also, apparently, categories on the company page currently have a `.no-link-declaration` class for some reason. Please remove that as well. All other issue will be solved by #428.

botex98:
@zner0L i've done the 2 modifications. 
I've encountered 1 more 'problem'. 
The article figure may be decorated depending if it's only an `image` or an image in `a` tag. 

Screenshot from 2020-10-06 12-16-41

botex98:
@zner0L Any updates?

zner0L (changes requested):
I'm sorry for keeping you hanging! I think you did good work! I don't like the decorated images, though. I can't see any good way of solving it, so I guess it's just enough if you add the `.no-link-decoration` class to the images in articles. You can do this in the shortcode, here: `layouts/shortcodes/img.html`.

botex98:
@zner0L done. :)

zner0L (approved):
Thanks a lot for contributing and putting up with my irregular reviewing schedule ;)

Issue #427, opened 2020-10-05
Wrap everything main in <main>


zner0L:
We basically wrap everything in main already. I'd like to keep it that way, especially since this becomes meaningful with #426. I don't know where this is needed (so you'd need to do some research), but one page I found was https://datarequests.org/id-data-controls.

jadonwolffs:
Can I give this a go for Hacktoberfest?

GoGi2712:
Can I work on this issue?

jonahcrowlinson:
> Can I give this a go for Hacktoberfest?

> Can I work on this issue?

@jadonwolffs @GoGi2712 - if you can fix this issue, then fix the issue. Unless it says different in the Code of Conduct or CONTRIBUTING.md there is no need to ask if you can fix it. Just make sure to link the issue in a pull request using #ISSUE_ID.

zner0L:
We'll, I hope you don't both started working on this. This is not a competition. I'll assign @jadonwolffs, as they were the first to ask. Also: asking is not needed, but appreciated @jonahcrowlinson.

PR #428, opened 2020-10-05
Added main tag to all pages

jadonwolffs:
As per issue #427 I went through all the pages of the site and added <main> tags to wrap the main content of the 2 pages that I found to not have it

I tried to use the fix and lint scripts to keep the formatting consistent but some things did change the first time I did it so I re-forked and added the tags again

### DIFF
--- a/src/id-data-controls.js
+++ b/src/id-data-controls.js
@@ -39,115 +39,117 @@ class IdDataControls extends preact.Component {
     render() {
         if (Privacy.isAllowed(PRIVACY_ACTIONS.SAVE_ID_DATA)) {
             return (
-                <div id="id-data-controls-container" className="narrow-page">
-                    <DynamicInputContainer
-                        key="id-data-controls"
-                        id="id-data-controls"
-                        onChange={this.handleCustomChange}
-                        fields={this.state.custom_id_data}
-                        title={t('saved-data', 'id-data-controls')}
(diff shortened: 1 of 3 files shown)


src/id-data-controls.js
```diff
@@ -39,115 +39,117 @@ class IdDataControls extends preact.Component {
...
-                                            </label>
-                                        </strong>
-                                    </div>
-                                    <div className="col60" style="padding-left: 10px; box-sizing: border-box;">
-                                        <InputControl
-                                            id="name-input"
-                                            suffix="fixed-id-data"
-                                            onChange={e => this.handleFixedChange('name', e)}
-                                            value={this.state.fixed_id_data['name']}
+                <main>
```
zner0L, line 42: How about adding the main tag just at the end here:

https://github.com/datenanfragen/website/blob/f07be204fdc5663934cdb32b19679f423b7cbbc9/src/id-data-controls.js#L241-L247

zner0L (changes requested):
Thanks for your contribution! Take a look at the comments and then I am happy with merging :)

src/id-data-controls.js
jadonwolffs: That's a much better way of doing that, I've made the changes I think

zner0L (approved):
Thanks!

PR #440, opened 2020-10-11
Only set changed_saved_companies cookie if the user actually changed something

### LINKED ISSUE #433: Only set changed_saved_companies cookie if the user actually changed something
In the wizard on the homepage, we use the `changed_saved_companies` cookie to determine whether the user has changed the default suggested companies (to test, use Germany, Austria or the UK as the country; we only have suggested companies for them). However, this cookie is unnecessarily set whenever the user enters the homepage.

Instead, we should set the cookie only if the user actually changed something. The cookie should not be set if the user just loads the homepage. In addition, it should instead be deleted whenever we currently set it to `false`.

The relevant code is in this file: https://github.com/datenanfragen/website/blob/master/src/Components/Wizard.js

alejandrotoga02:
# Only set changed_saved_companies cookie if the user actually changed something

fix #433

### DIFF
--- a/src/Components/Wizard.js
+++ b/src/Components/Wizard.js
@@ -238,7 +238,7 @@ export class SavedCompanies {
     }
 
     getUserChanged() {
-        return Cookie.get(USER_CHANGED_COOKIE) && true;
+        return Cookie.get(USER_CHANGED_COOKIE) ? true : false;
     }
     setUserChanged(value = true) {
         if ( value ){
--- a/src/Components/Wizard.js
+++ b/src/Components/Wizard.js
@@ -244,7 +244,7 @@ export class SavedCompanies {
         if ( value ){
             Cookie.set(USER_CHANGED_COOKIE, value, { expires: 365 });
         } else {
-            Cookies.remove(USER_CHANGED_COOKIE);
+            Cookie.remove(USER_CHANGED_COOKIE);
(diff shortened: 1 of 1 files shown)


src/Components/Wizard.js
```diff
@@ -238,10 +238,14 @@ export class SavedCompanies {
     }
 
     getUserChanged() {
-        return Cookie.get(USER_CHANGED_COOKIE) === 'true';
+        return Cookie.get(USER_CHANGED_COOKIE) ? true : false;
```
baltpeter, line 241: You should be able to just convert to a boolean like this:

```suggestion
        return !!Cookie.get(USER_CHANGED_COOKIE);
```

baltpeter (changes requested):
Thanks for the changes, @alejandrotoga02! I have just left one small comment below.

Also, please make sure that you run Prettier with our configuration of the file to make sure it matches our formatting style. If you follow the instructions in the README, this should happen automatically on committing.

alejandrotoga02:
@baltpeter  okok  I'm going to do the changes :D

src/Components/Wizard.js
alejandrotoga02: done

baltpeter (approved):
Great, thanks again.

Issue #422, opened 2020-10-02
Nicer event titles in the ics export


mal-tee:
The .ics export is a very handy feature, but I like to have an overview over my calendar without clicking on events.
In this case the title could be more precise:

![grafik](https://user-images.githubusercontent.com/45886910/94941503-3f364b00-04d5-11eb-99f3-d7b1f1dd6753.png)
![grafik](https://user-images.githubusercontent.com/45886910/94941547-4eb59400-04d5-11eb-97c5-391d8b22306b.png)

As you can see there is only one request due.

I would like to have the companies name in the title of the event (this is called `SUMMARY` in the .ics). This might even work for a few companies at once (depending on the length?). For more we could show a count in the title.

This is where the magic happens:

https://github.com/datenanfragen/website/blob/1aed9ca5644e92e90d2a3bbf4f1593d16348f497/src/my-requests.js#L361

king-11:
Hey @mal-tee can I start working on this?

mal-tee:
Sure, go ahead!

king-11:
@mal-tee sorry for the delay but am not able to test the website locally the docs aren't very clear can you help out a bit, please

king-11:
thanks a lot @mal-tee I was really confused now I will start working on this issue you can hope for a pr soon will try my best

mal-tee:
Great, have fun coding 👍

king-11:
@mal-tee so shall I pass in instead of 'ics-summary' as new variable whose value will be set as per the number of requests if I guess 2 companies is good so if we have two companies then there recipients else the count how does that sound .

https://github.com/datenanfragen/website/blob/1aed9ca5644e92e90d2a3bbf4f1593d16348f497/src/my-requests.js#L361

mal-tee:
Sounds good

PR #442, opened 2020-10-12
nicer event titles

king-11:
nicer event titles display a few company names else show count
﻿regex split use to remove access token as is not necessary for summary


king-11:
@mal-tee review please closes issue #422

mal-tee:
@baltpeter Can you review this please? :D

src/my-requests.js
```diff
@@ -349,6 +349,14 @@ class RequestList extends preact.Component {
...
                 );
             });
+            const re = /\(.+\)$/;
+            let heading =
+                items.length <= 2
+                    ? `for ${items
+                          .map((item) => item.split(re)[0].trim())
+                          .slice(0, 2)
+                          .join(', ')}`
+                    : `for ${items.length} companies`;
```
baltpeter, line 359: Instead of just showing the number here, it would be great to still list first one or two company names and then say "and <n> more companies" or something like that.

baltpeter (changes requested):
Thanks for the PR! I have left some comments below.

src/my-requests.js
```diff
@@ -349,6 +349,14 @@ class RequestList extends preact.Component {
                     } – ${request.reference})`
                 );
             });
+            const re = /\(.+\)$/;
```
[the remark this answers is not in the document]
king-11: okay resolved

src/my-requests.js
king-11: okay done

king-11:
@baltpeter  rest i have resolved this translation part am not getting a bit; the current changes are as follows 
```js
grouped_requests[group].forEach((id) => {
                let request = this.state.requests[id];
                titles.push(request.recipient.split('\n', 1)[0])
                items.push(
                    `* ${request.recipient.split('\n', 1)[0]} (${
                        request.type === 'custom' && request.response_type
                            ? t(request.response_type, 'generator')
                            : t(request.type, 'my-requests')
                    } – ${request.reference})`
                );
            });
            titles = titles.map((title) => title.length > 10 ? `${title.slice(0,10)}..` : title)
            let heading = ` for ${titles.slice(0,2).join(" ,")}`
            heading += titles.length > 2 ? ` and ${titles.length - 2} more` : ""
```

how can I use t function for this I am bit not sure could you help out a bit

baltpeter:
The translation files are in [this directory](https://github.com/datenanfragen/website/tree/master/src/i18n). For example, the English translations for the 'My requests' feature are here:

https://github.com/datenanfragen/website/blob/ce425600cca7111ba1bf187b165462f6536a3352/src/i18n/en.json#L229-L258

You can simply add new translations to that object and then use them like this: `t(your_translation_key, 'my-requests')`

king-11:
@baltpeter so there are three words am using 'for' 'more' and 'and' so i have to add them into the file.

baltpeter (changes requested):
Thanks for making the changes, this is almost done now. You don't need to add the translations for the other languages (that will be done by our contributors on Weblate) but currently, even the English translations aren't displaying correctly. This is because you also need to pass the context to the `t()` function like this: `t('<translation_key>', 'my-requests')`

king-11:
> Thanks for making the changes, this is almost done now. You don't need to add the translations for the other languages (that will be done by our contributors on Weblate) but currently, even the English translations aren't displaying correctly. This is because you also need to pass the context to the `t()` function like this: `t('<translation_key>', 'my-requests')`

ohh missed that will get in done in just 5 mins

king-11:
@baltpeter i guess finally its done :smile:

baltpeter (approved):
Thanks again.

king-11:
@baltpeter @mal-tee thanks for helping out throughout 😄 felt great to make contribution to this really amazing project.

mal-tee:
In case you haven't seen it: We also have our own little hacktoberfest-thingy: https://www.datarequests.org/blog/hacktoberfest-2020/

PR #446, opened 2020-10-13
Add stub file creation for countries with no data.

### LINKED ISSUE #434: Wizard: Can fetch 404 error be suppressed?
The wizard on our homepage loads the suggested companies from a URL of the form: `https://<domain>/db/suggested-companies/<country_code>_wizard.json`

If we don't have suggested companies for a country (currently, we only have them for DE, AT and UK), this request fails with a 404 error. This behaviour is expected and handled in the code (in this case, we just don't show any suggested companies).

However, browsers still show an error message in the console. To reproduce this, visit the homepage and set the country to `all`. Notice the following error in the browser console:

image

Can this error be suppressed? I didn't find a way from a quick look at MDN but I may well have missed something. It is however also possible that this cannot be fixed.

This is the relevant code:

https://github.com/datenanfragen/website/blob/05133ee1ee4d147c1cf643cde8161291e300332c/src/Components/Wizard.js#L62-L79

rgroves:
In the prepare-deploy script, countries with no data now have stub files created so that the wizard no longer generates a 404 in the browser when fetching the data for those countries.

Closes #434.

### DIFF
--- a/prepare-deploy.js
+++ b/prepare-deploy.js
@@ -1,20 +1,41 @@
 const glob = require('glob');
 const fs = require('fs');
 const path = require('path');
+const schema = require('./static/schema');
 
-glob('static/db/suggested-companies/??.json', async (err, countries) => {
-    countries.forEach(country_file => {
+// Find countries that have data.
+let do_not_stub_countries = [];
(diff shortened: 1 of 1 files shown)


baltpeter (approved):
Thank you!

PR #451, opened 2020-10-14
Highlight tested and unverified companies in search results and company single pages

baltpeter:
This needs to be merged before datenanfragen/data#773.


Sakshamoo17:
layouts/company/single.html

Sakshamoo17:
layouts/company/single.html
layouts/company/index.html

src/Components/SearchBar.js
```diff
@@ -71,25 +71,34 @@ if (Privacy.isAllowed(PRIVACY_ACTIONS.SEARCH)) {
...
 
-                                let name_hs = suggestion.highlights.filter(a => a.field === 'name');
-                                let runs_hs = suggestion.highlights.filter(a => a.field === 'runs');
+                                let name_hs = suggestion.highlights.filter((a) => a.field === 'name');
+                                let runs_hs = suggestion.highlights.filter((a) => a.field === 'runs');
 
                                 return (
                                     '<span><strong>' +
                                     (name_hs.length === 1 ? name_hs[0].snippet : d.name) +
+                                    (d.quality === 'tested'
```
mal-tee, line 93: FF a11y checking tool doesn't like this: 
grafik

mal-tee (changes requested):
Thank you very much! Such a cool feature. I've left a small comment.

baltpeter:
Interesting. Thanks for checking. The icon really shouldn't be clickable itself. I'll have to look into that.

baltpeter:
Genuinely not sure how we could fix this. The [documentation for the issue](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Keyboard#Clickable_elements_must_be_focusable_and_should_have_interactive_semantics) says that this applies to elements that "can be clicked with a pointing device" and "An element is clickable if it has an `onclick` event handler defined."

The icon doesn't have an `onclick` handler. I didn't find a way to communicate to Firefox that this element isn't clickable (itself). Setting `role="figure"` or `role="img"` doesn't change anything. Neither does setting `pointer-events: none;` (this also breaks the title text, so not an option anyway). Setting `tabindex="0"` on option div (which is actually supposed to be clickable) doesn't help either.

I wasn't even able to get rid of the issue by setting `tabindex="0"` on the icon itself as the issue description suggests (this also isn't actually an option as that would make the accessibility of the search bar worse imo). 

So yeah, no idea what to do here. For reference: axe doesn't complain about that.

mal-tee (approved):
Thank you for looking into this! Seems more like it's a problem on Firefox' side, especially if axe doesn't complain. I say: Let's merge and recheck this with future FF (and axe) versions.

PR #452, opened 2020-10-14
Improve sorting on company database page.

### LINKED ISSUE #245: Bring accentued letters together on company database page
I noticed that there were some companies starting with Ö but they are displayed after Z.

I guess it could happen with other accentued letters. I think each accentued letter should be interpreted as the normal one. You could use the following correspondance table (non exhaustive).

```
A = Å, Á, À, Â, Ä, Ã
E = É, È, Ê, Ë
I = Ì, Í, Î, Ï
O = Ò, Ó, Ô, Õ, Ö, Ø
U = Ŭ, Ù, Ú, Û, Ü
Y = Ý, Ÿ
C = Ç
N = Ñ
```

rgroves:
This change will make it so that accented characters in the company name are treated as equivalent Latin Alphabet characters in the sorted list on the company database page.

The change also filters out certain special characters, like various quotes, and problem characters, like Zero Width Space, so that they do not affect sorting.

Closes #245

### DIFF
--- a/layouts/company/list.html
+++ b/layouts/company/list.html
@@ -13,7 +13,30 @@
     <!-- All of this only to make case-insensitive sorting possible. I hate you hugo -->
     {{ $pages := (.Data.Pages.ByParam "name")}}
     {{ range $pages }}
-        {{ $.Scratch.SetInMap "upper_title" (upper .Params.name) .Params.slug }}
+        {{/*
+             Normalize company name for sorting by uppercasing and translating accented characters to equivalent Latin
+             alphabet characters.
+        */}}
+        {{ $clean_name := replaceRE "[ÅÁÀÂÄÃ]" "A" (upper .Params.name) }}
+        {{ $clean_name := replaceRE "[Ç]" "C" $clean_name }}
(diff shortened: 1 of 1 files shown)


rgroves:
Just wanted to note an interesting case. I took the screen shot below from the company page of the current live website. You'll notice the entry for "Adfire Health LLC" is in there with some of the others that had special/accented letters as the first character in the name. 

Once I started testing my change to translate the accented characters to equivalent Latin characters, it seemed to be working except for "Adfire Health LLC." I was perplexed as to why that one was still sorting at the end of the list. After some inspecting in the browser, I came to realize that it started with a this character: [U+200B ZERO WIDTH SPACE](https://util.unicode.org/UnicodeJsps/character.jsp?a=200B).

I just wanted to call this case out, in case it makes sense to you to add some data cleansing/validation when accepting info on new companies. With the replacements happening in the template now it may not be a huge deal to just add some more characters there, but I think it makes sense to do some cleanup at the time the data is collected if possible. Just something to consider.

image

mal-tee:
Oh, thats very interesting. Nice find.

baltpeter (approved):
That is a very nice implementation. Thank you very much, @rgroves!

Also thanks for pointing out the

PR #461, opened 2020-10-16
Placeholder for related companies in the dev build #367

### LINKED ISSUE #367: Placeholder for related companies in the dev build
Atm the company pages do not contain related companies in dev build. I think we should add placeholders.

chrysillala:
Fix #367
By adding else condition to display the placeholder content

### DIFF
--- a/layouts/company/single.html
+++ b/layouts/company/single.html
@@ -135,6 +135,9 @@
         <ul>
             <li><a href="{{ ref . "company/datenanfragen" }}">Darkenanfragen AG</a></li>
             <li><a href="{{ ref . "company/datenanfragen" }}">Acme Inc.</a></li>
+            <li><a href="{{ ref . "company/datenanfragen" }}">Ace Corporation</a></li>
+            <li><a href="{{ ref . "company/datenanfragen" }}">Foo Bar Co.</a></li>
+            <li><a href="{{ ref . "company/datenanfragen" }}">Mum and Pop Ltd.</a></li>
         </ul>
     </div>
     {{ end }}
--- a/layouts/company/single.html
+++ b/layouts/company/single.html
@@ -129,6 +129,17 @@
                 </ul>
             </div>
         {{ end }}
+    {{ else }}
+    <div id="related-companies">
+        <h2>Related companies</h2>
+        <ul>
+            <li><a href="https://www.datarequests.org/company/quotient-com/">Quotient Technology Inc.</a></li>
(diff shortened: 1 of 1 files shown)


layouts/company/single.html
```diff
@@ -129,6 +129,17 @@ <h2>{{ T "cdb-related-companies" }}</h2>
                 </ul>
             </div>
         {{ end }}
+    {{ else }}
+    <div id="related-companies">
+        <h2>Related companies</h2>
```
baltpeter, line 134: This needs to be translated, just like we do in the "real" version.

layouts/company/single.html
```diff
@@ -129,6 +129,17 @@ <h2>{{ T "cdb-related-companies" }}</h2>
...
+        <ul>
+            <li><a href="https://www.datarequests.org/company/quotient-com/">Quotient Technology Inc.</a></li>
```
baltpeter, line 136: After some internal discussion, we think that it would be better to use obvious fake company names (something like *Acme Inc.*) here to make it clear that these are only placeholders. These should then all link to our company page. Please use `{{< ref "company/datenanfragen" >}}` to make sure the URL is currently localized.

baltpeter (changes requested):
Thanks for the PR! I have left some comments below.

layouts/company/single.html
mal-tee: You could use "Darkenanfragen AG"! 😄

chrysillala:
> Thanks for the PR! I have left some comments below.

Thanks! Will update according to your feedback.

chrysillala:
@baltpeter @mal-tee just update, kindly recheck my PR. Thanks!

mal-tee:
Thank you! However, I think there should be just as many placeholders as real links, i.e. 5.

chrysillala:
> Thank you! However, I think there should be just as many placeholders as real links, i.e. 5.

@mal-tee Noted. Do you have any preferences for the other 3 placeholder links? Or should I just duplicate the previous one I created until there are 5 links?
Thanks!

mal-tee:
If you can come up with a few other fake (funny) companies go ahead and add them. Otherwise you can duplicate the other ones. :D

chrysillala:
Hi, please recheck my PR, just add more placeholders. Hopefully this one can solve the issue. Thank you!

mal-tee (approved):
LGTM

baltpeter (approved):
Thanks!

PR #464, opened 2020-10-18
Translate legal notice into Portuguese

ShyamGit01:
The key contents are translated into Portuguese. If it is in the correct format feel free to give a mail so that I will complete all the files.

### DIFF
--- a/content/en/contact.md
+++ b/content/en/contact.md
@@ -11,20 +11,20 @@ Schreinerweg 6
 38126 Braunschweig  
 Germany
 
-**Phone:** +49 531 209299 35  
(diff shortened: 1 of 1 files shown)


baltpeter (changes requested):
Thanks for the translation, @ShyamGit01! Instead of modifying the existing file, please create a new file `content/pt/contact.md`—this way, the content is available in both English and Portuguese.

ShyamGit01:
Yes I am interested for it, please assign me to do so in the respective folders.

baltpeter:
No need to assign you, you can just put the file in the folder. :)

PR #465, opened 2020-10-19
Optimize images

JeffResc:
This pull request optimizes the images found in the repository using lossless compression. This means that the image quality will not be degraded. This will mean less space taken up by images along with faster loading times when viewing them. The overall file size has been reduced by around 5%.


baltpeter:
Thanks for the PR. Our [commit pipeline](https://github.com/datenanfragen/website/blob/4c13bd5ad71b3cac309d4863da2b7a4de3e83df6/package.json#L127) already optimizes images, which is why you only got minimal improvements from optimizing them again.

Issue #432, opened 2020-10-10
Privacy controls: Only set cookie if preference != default


baltpeter:
We are saving the user preferences for our [privacy controls](https://www.datarequests.org/privacy-controls) as cookies. Currently, we always set the cookie if the user changes something. However, this isn't necessary.

If a preference is set back to the default value, we can just delete the cookie.

The relevant code is here:

https://github.com/datenanfragen/website/blob/05133ee1ee4d147c1cf643cde8161291e300332c/src/Utility/Privacy.js#L39-L41

Note how the default value is determined:

https://github.com/datenanfragen/website/blob/05133ee1ee4d147c1cf643cde8161291e300332c/src/Utility/Privacy.js#L35

ayush015:
@baltpeter Can I take a look ?

baltpeter:
Sure, @ayush015. I'll assign you.

ayush015:
@baltpeter how can I set it up ? if you can help me that'd be great

abhishekUpmanyu:
@baltpeter Hi! I have made the necessary change, can you please assign this to me, I'm looking forward to contribute

baltpeter:
@ayush015 Are you still working on this? If so, we'll wait for your PR. Otherwise, I'll review the one by @abhishekUpmanyu.

ayush015:
nope I way busy somewhere else .

Issue #429, opened 2020-10-06
Fix intermittent test failures in old-bugs/country-change-updates-wizard test


baltpeter:
This is the test in question:

https://github.com/datenanfragen/website/blob/2669b17b66859ee4e20cfd3c145708d442f5daa7/cypress/integration/old-bugs/country-change-updates-wizard.spec.js#L30-L57

In particular, the test fails when verifying the cookie value:

https://github.com/datenanfragen/website/blob/2669b17b66859ee4e20cfd3c145708d442f5daa7/cypress/integration/old-bugs/country-change-updates-wizard.spec.js#L44

[This](https://app.circleci.com/pipelines/github/datenanfragen/website/1042/workflows/365d6a57-64b7-48e7-af98-18c5177b788e/jobs/1699) is a CI log of a failed run (the log is public but you unfortunately need to be logged into CircleCI to view it).

[These](https://dashboard.cypress.io/projects/nug7wk/runs/441/specs) are the details for the same failed run in the Cypress dashboard.

We already tried to add a wait. This doesn't help, though, and we would like to avoid that if possible:

https://github.com/datenanfragen/website/blob/2669b17b66859ee4e20cfd3c145708d442f5daa7/cypress/integration/old-bugs/country-change-updates-wizard.spec.js#L41

See [this README](https://github.com/datenanfragen/website/blob/master/cypress/README.md) for more details on how our tests work.

jadonwolffs:
I was going to try work on this but it appears to pass fine on master for me already when testing using Electron 80 but seems to fail with Firefox 81. The [README](https://github.com/datenanfragen/website/blob/master/cypress/README.md) mentions that it should be running in Electron on the CI env (and it says that it is running `Electron 80 (headless)` on [circleci](https://app.circleci.com/pipelines/github/datenanfragen/website/1042/workflows/365d6a57-64b7-48e7-af98-18c5177b788e/jobs/1699)) so I'm not sure what the difference between me and the CI env is

Maybe it's something different with my setup? (WSL Ubuntu 20.10 with Xserver on Windows 10)

## Electron 80:
![image](https://user-images.githubusercontent.com/16850875/95190943-52e0fa80-07d0-11eb-98aa-b4df3411010d.png)
## Firefox 81:
![image](https://user-images.githubusercontent.com/16850875/95191475-1feb3680-07d1-11eb-9ba1-1570e3b8c7a0.png)

baltpeter:
Thanks for looking into this!

[...]

PR #467, opened 2020-10-20
Update Privacy.js to only set cookie if preference != default

abhishekUpmanyu:


src/Utility/Privacy.js
```diff
@@ -37,7 +37,11 @@ export default class Privacy {
     }
 
     static setAllowed(privacy_action, value) {
-        Cookie.set(this.cookieNameForAction(privacy_action), value, { expires: 365 });
+        if(value !== privacy_action.default) Cookie.set(this.cookieNameForAction(privacy_action), value, { expires: 365 });
+        else {
+            let cookie_value = Cookie.get(this.cookieNameForAction(privacy_action));
+            if(cookie_value!==undefined) Cookie.remove(this.cookieNameForAction(privacy_action));
```
baltpeter, line 43: Are you sure that this is necessary? I am pretty sure that you can just delete the cookie, even if it doesn't exist.

baltpeter (changes requested):
Thanks for the PR, @abhishekUpmanyu! I have left one small comment below.

Two more things:

* Please make sure that you run Prettier on the file to make sure it complies with our code style. This should happen automatically when committing if you followed the instructions in our README.
* As you can see from the [CI results](https://app.circleci.com/pipelines/github/datenanfragen/website/1168/workflows/727eeb1e-9759-4293-bf1e-c5d2d3d4b884/jobs/1919), the `privacy-controls/toggle.spec.js` test is now failing as it assumes the previous behaviour. This needs to be changed before we can merge. See [this](https://github.com/datenanfragen/website/blob/master/cypress/README.md) on how the tests work.

src/Utility/Privacy.js
abhishekUpmanyu: Oh, didn't think about that, I'll make the necessary change

abhishekUpmanyu:
@baltpeter I've made the changes, and have been trying to start hugo server for running cypress tests, but I get this error while trying to run the command `hugo server`

```
Start building sites … 
ERROR 2020/10/22 14:57:24 Alias "de/e.v." contains component with a trailing space or period, problematic on Windows
Built in 47479 ms
Error: Error building site: cannot create "de/e.v.": Windows filename restriction
```

Any ideas how to get past this?

baltpeter:
@abhishekUpmanyu Interesting. Apparently filenames may not end with a dot on Windows, surprisingly you're the first one to report that problem.

You can just delete the `"e.v."` aliases in `content/*/verein/_index.md`. That should allow the site to build.

abhishekUpmanyu:
I've made the changes and the test that was changed now runs successfully but now I have another test failing, maybe because the cookie is not set. Please guide me a bit here

baltpeter:
Thanks, the changes look good.

Not sure about the test failure. I was not able to reproduce that on my machine, here the test passes fine. Does the test fail when you run it locally? We are having some intermittent failures with this test (see #429) but this seems to be different. I'll try running the test a couple of times in CI to see if it is also intermittent.

abhishekUpmanyu:
To be honest, there were issues trying to get this running on my machine, so I basically coded and pushed. I see that the tests have passed now. Will take a look at the issue you've mentioned. Cheers!

abhishekUpmanyu:
This pull request fixes #432

baltpeter:
Ok. Well, the problem seems to have been only temporary. The test has passed several times in a row now, so I think we can merge.

baltpeter (approved):
Thanks again.

Issue #402, opened 2020-09-19
MailtoDropdown: Support for more webmailers


baltpeter:
Our MailtoDropdown doesn't just support the regular `mailto:` 'protocol' but also various webmailers (currently Gmail and Yahoo):

![Peek 2020-09-19 10-02](https://user-images.githubusercontent.com/4048809/93662287-90aef680-fa4e-11ea-85ec-679bd465f0cb.gif)

We would like to extend this with support for more webmailers, especially local ones like GMX and Web.de in Germany.

The handlers are implemented here:

https://github.com/datenanfragen/website/blob/6d6fa4b7bda928fd9671654f604ab8b672ba694e/src/Components/MailtoDropdown.js#L15-L100

Adding support for a new one is easy once you have found the correct URL.

raj5036:
# wjhewuqeuhwqjhkqwje

haardikdharma10:
Hi @baltpeter, I'd like to trigger this and add GMX and Web.de. Can I go ahead?

baltpeter:
@haardikdharma10 Sure, go ahead!

galanggg:
hello @baltpeter can i send a PR for link of Outlook ?

swapniljha001:
> hello @baltpeter can i send a PR for link of Outlook ?

@galanggg  The Outlook wasn't working previously, but the reference material is 4 years old, so you might need to research some more before adding the Outlook link.

```
     // TODO: Outlook doesn't work. I think our body may be too long. :( 
     // Reference: https://blogs.msdn.microsoft.com/carloshm/2016/01/16/how-to-compose-a-new-message-or-event-and-populate-fields-in-office365/ 
     // outlook: { 
     //     link: d => 
     //         `https://outlook.live.com/owa/?path=/mail/action/compose&to=${d.email}&subject=${d.subject}&body=${d.body}`, 
     //     countries: ['all'] 
     // }, 
```

swapniljha001:
What other web mailers are needed here? I think the most common ones like Google, Yahoo, Yandex have already been added to this.

galanggg:
@swapniljha001 heres my PR #416

baltpeter:
> What other web mailers are needed here? I think the most common ones like Google, Yahoo, Yandex have already been added to this.

[...]

PR #473, opened 2020-10-26
added support for AOL webmailer

privateOmega:
For #402

### DIFF
--- a/src/Components/MailtoDropdown.js
+++ b/src/Components/MailtoDropdown.js
@@ -14,12 +14,12 @@ let previous_active_element = document.body;
 // to produce links for GMX and Web.de.
 export const MAILTO_HANDLERS = {
     mailto: {
-        link: d => `mailto:${d.email}?subject=${d.subject}&body=${d.body}`,
-        countries: ['all']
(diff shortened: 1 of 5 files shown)


privateOmega:
@baltpeter PTAL

baltpeter:
Thanks for the PR, @privateOmega. AOL is a good addition.

Unfortunately, all line breaks were lost when I tested the feature. This needs to be fixed before we can merge.

image

privateOmega:
@baltpeter Sure I will take a look at it when I get some time. And will try to resolve it ASAP.

privateOmega:
@baltpeter I guess this needs to be put on hold indefinitely.
I tried different types of line feed and carriage return URL encoding like %0A,%0D, %0D%0A etc, but nothing seems to work and I guess at AOL mail side, they have an issue in their function which decodes the URL component.

baltpeter:
That's unfortunate. Thanks for trying, though! I have still applied the 'hacktoberfest-accepted' label for your effort. :)

privateOmega:
@baltpeter Thanks a lot.

PR #476, opened 2020-10-31
Add french overview

eliottha:
Hey hey,

I added the french quick overview. 
Please keep me posted about it!


baltpeter (commented):
Thank you very much for the translation, @eliottha! I'll wait for your reply on #477 before merging.

Other than that, the translation looks great. Just two minor comments:

* You accidentally left one English line in. That should be deleted.
* I don't think it's necessary to copy the image to the French translation. Hugo should share the images between translations. The German translation doesn't have the image either.

eliottha:
That line is translated just above just forgot to remove the english line. Also changing to informal form now. Leave #477 open, instead of making everything formal, I'll change all to informal

eliottha:
@baltpeter should be fine now

baltpeter:
Thanks for making the changes!

Sorry, I didn't pay enough attention: The blog post needs to have the same file/folder name as in the English version, otherwise Hugo won't recognize that it is a translation (which means that the language chooser doesn't work correctly and the images aren't shared). The URL is already set to the French translation by the `slug` property. Usually, we also specify an alias so that the English URL also works.  
I have fixed that in 7535a4e6c9556ca863fa7f969ca2de1ecba3ac18.

eliottha:
Thanks. So it can be merged now I guess?

baltpeter:
Yep, just waiting for the build and test to finish. :)

baltpeter (approved):
Thanks again!

eliottha:
> Thanks again!

You're more than welcome. 5 more to go for now!

PR #482, opened 2020-11-02
Fixes #480: Properly label inputs for the ActWidget

### LINKED ISSUE #480: Inputs in ActWidgets are not screen reader-friendly
In the ActWidget, the form labels aren't set correctly. They refer to hidden inputs that don't actually do anything:

image

This is because the ActWidget inherits from the generator where those hidden inputs are actually visible and allow the user to change the field description themselves.

baltpeter:


mal-tee (approved):
Thank you very much!

PR #508, opened 2020-11-15
Add test for /suggest

mal-tee:


cypress/integration/use-cases/suggest.spec.js
```diff
@@ -0,0 +1,46 @@
...
+        cy.contains('You can follow the progress')
+            .should('have.attr', 'href')
+            .and('match', /MAGICSTRING/);
+        cy.contains('You can follow the progress').clickLinkWithoutFollowingHref();
+
+        // now check if the fetchlog contains the suggestion
+        cy.window().then((win) => {
+            const body = win.fetchlog[0][1].body;
+            const json = JSON.parse(body);
+            if (json.data.name !== 'Test Company LLC') throw new Error('Name wrong.');
```
baltpeter, line 41: I'm pretty sure that this could also be implemented with `expect`/`should` which seems cleaner to me.

baltpeter (changes requested):

cypress/README.md
```diff
@@ -8,7 +8,7 @@ The tests are automatically run by the CI environment.
 
 To run them locally, you can use the Cypress test runner. Make sure you have the correct dependencies installed by running: `yarn cypress verify`
 
-If you want to, you can override the launch URLs:  
+If you want to, you can override the launch URLs:
```
baltpeter, line 11: You need to disable trailing whitespace trimming for Markdown. :D 

In VS Code I use:

```json
"[markdown]": {
    "files.trimTrailingWhitespace": false,
    "editor.formatOnSave": false
}
```

baltpeter (approved):
Awesome, thanks!

PR #511, opened 2020-11-20
Fixes #348: Properly handle Germany in SvaFinder + tests for the component

### LINKED ISSUE #348: Clarify Germany in SvaFinder
Currently the flow is:

1. "Where do you live?" -> Germany
2. "Which of the following areas does your request concern?" -> "Private entity"
3. "In which state is the entity your request concerns based?"

There needs to be a step between 2. and 3. that asks "Is the company also based in Germany?".

baltpeter:


mal-tee (approved):
Thank you very much! 
I have two small notes for additions, these are non-blocking but would be very much appreciated:

- Test the SvaWidget in the context of a complaint
- Let the German/other test check for the question "In which state do you live?"

mal-tee (approved):
Great, thanks!

PR #513, opened 2020-11-25
Skip components/svafinder/generator-complaints test in prod

baltpeter:
window.accessLocalForageStore is not available in production.


mal-tee (approved):

mal-tee:
Thanks!

PR #518, opened 2020-12-09
Upgrade cypress to version 6

mal-tee:
- Seems like only one test had to be changed, see https://docs.cypress.io/guides/references/migration-guide.html#Assertions-on-non-existent-elements
- `country-change-updates-wizard.spec.js` seems to be kinda flaky, at least on my system. Was that the case before too?
- the linux dependencies changed, see https://docs.cypress.io/guides/references/migration-guide.html#Linux-dependencies
  - worked out of the box for my machine
- someday in the future my latest suggest test could be rewritten to leverage the fact that cypress now supports interception of `fetch`-requests


mal-tee:
We could consider enabling retries to mitigate (potentially) flaky tests: https://docs.cypress.io/guides/guides/test-retries.html#Introduction

baltpeter:
I'll fix the merge conflict.

> `country-change-updates-wizard.spec.js` seems to be kinda flaky, at least on my system. Was that the case before too?

Yep: #429.

> We could consider enabling retries to mitigate (potentially) flaky tests: https://docs.cypress.io/guides/guides/test-retries.html#Introduction

I'm a little torn on that. Generally, I don't like the idea as I think that encourages an attitude of not caring about flakiness ("it'll just retry anyway…"). In the case of the before-mentioned tests, though we have spent quite a bit of time trying to fix the flakiness but haven't been successful, so I think it's a good idea to enable retries just for that test.

baltpeter (approved):
Awesome, thank you!

PR #523, opened 2020-12-19
Disable Percy for now

baltpeter:
It's currently broken anyway (#369) and since #518, the scheduled production test keeps failing (which seems to be related to Percy).


mal-tee (approved):
Thanks!

PR #526, opened 2021-01-14
Fix featured_image_url in two German pages

baltpeter:
People keep t(wee|oo)ting about us and the cards didn't work.


mal-tee (approved):
Thanks!

PR #537, opened 2021-01-24
Add request form spanish translations

fm:
Certain fields were showing up as 'null' when sending a request whose preferred language was 'es' 

Translations were successfully tested locally on dev environment


baltpeter (approved):
Great, thank you!

PR #547, opened 2021-02-02
update last_edited for index pages

mal-tee:
Otherwise the page looks a bit old in certain contexts, e.g.:

image

Notice how it says "2019". :D

We might consider doing this automatically, but I am not sure if it's worth the effort.


baltpeter:
Thanks for submitting the PR. I agree that having this old of an updated date looks off-putting. However, I am not a fan of manually specifying a date to be honest. For one, we are very likely to forget doing this regularly (as evidenced by the fact that we last did this in 2019).  
But more importantly, this feels disingenuous. We didn't actually change anything of substance yesterday, so I don't see why we would specify that as the updated date.

If, instead we simply specify neither `date` nor `last_edited`, [we will fallback to Hugo's `.Date`](https://github.com/datenanfragen/website/blob/3cb3663f6edb11bed28c317224f8a29f1156bee3/layouts/partials/head/meta.html#L1-L9). From a bit experimenting, this seems to use the date of the last post published for the respective language. That seems like a much more sensible and honest date to me, with the added benefit of less overhead for us.

mal-tee:
I agree that this manual tweaking feels kinda disingenuous. But one could argue that we update `data` almost daily, thus the page as well. 

But I agree, using the last published post is cleaner and more comfortable. :D

baltpeter:
> But one could argue that we update data almost daily, thus the page as well.

Sure, but then we would have to actually update the date daily as well. :D

baltpeter (approved):
Thank you!

PR #548, opened 2021-02-02
 Fixes #538: Add "child-src 'self'" to CSP to fix PDF download in Safari 

### LINKED ISSUE #538: CSP error prevents PDF download from loading on Safari
**OS & Browser details:**
macOS Big Sur 11.1 (20C69)
Safari 14.0.2 (16610.3.7.1.9)

Screen Shot 2021-01-24 at 4 06 19 PM

Could not replicate on Chrome nor Firefox

baltpeter:


baltpeter:
Seems like Safari doesn't understand the [`worker-src`](https://developer.mozilla.org/en-us/docs/Web/HTTP/Headers/Content-Security-Policy/worker-src#browser_compatibility) directive that we were relying on for the PDF worker but instead expects [`child-src`](https://developer.mozilla.org/en-us/docs/Web/HTTP/Headers/Content-Security-Policy/child-src#browser_compatibility) to be set for this purpose.

This PR fixes that on Catalina in my sosumi VM. @fm, can you confirm this on your machine as well?

fm:
@baltpeter worked in Big Sur

PR #550, opened 2021-02-09
Suggest: Remove 'null's from arrays, trim strings

mal-tee:


mal-tee:
github 🙄

mal-tee:
We will move this into https://github.com/datenanfragen/backend.

PR #562, opened 2021-03-21
Respect prefers-reduced-motion

### LINKED ISSUE #531: Respect prefers-reduced-motion for the typewriter effect on the homepage

davenewham:
Closes #531

### DIFF
--- a/src/home.js
+++ b/src/home.js
@@ -16,8 +16,7 @@ render(<Wizard />, wizard_div.parentElement, wizard_div);
 
 function typewriter(text, i, fnCallback) {
     if (text && i < text.length) {
-        document.getElementById('home-hero-word').innerHTML =
-            text.substring(0, i + 1) + '<span aria-hidden="true"></span>';
+        document.getElementById('home-hero-word').textContent = text.substring(0, i + 1);
 
         setTimeout(function () {
             typewriter(text, i + 1, fnCallback);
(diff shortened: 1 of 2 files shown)


src/home.js
```diff
@@ -39,5 +39,11 @@ function startTextAnimation(i) {
 }
 
 window.onload = () => {
-    startTextAnimation(0);
+    const reducedMotion = window.matchMedia('(prefers-reduced-motion: reduce');
```
baltpeter, line 42:
```suggestion
    const reducedMotion = window.matchMedia('(prefers-reduced-motion: reduce)');
```

baltpeter (changes requested):
Thanks, @davenewham!

There is currently (sometimes) a short flash of the first word before the JS kicks in:

ezgif-7-9b71bdfca345

Not sure what the best way to go about this is. Maybe we should just start off with no word in the HTML?

https://github.com/datenanfragen/website/blob/25b3fa755acce08764d6c0075e893583616e96da/layouts/index.html#L6

davenewham:
Good spot, I didn't notice that. Used a <br> tag to prevent the elements moving around (not sure if there's a better way), however because of that, people viewing with js turned off they will see an extra line between the button and the text.
image

baltpeter (approved):

baltpeter:
> however because of that, people viewing with js turned off will see an extra line between the button and the text.

I think that's fine. Thanks!

PR #564, opened 2021-04-08
Fix request language fallback

zner0L:


baltpeter:
Haven't looked at this in depth yet, but we should really add a test to check that we don't regress on this bug. Do you want to add that, @zner0L?

baltpeter:
Other than the missing test, this PR is approved.

baltpeter:
@zner0L I'm really not a fan of disabling web security, so here's an alternative solution that instead relies on the fact that different tests may in fact have different origins. If you're OK with that, we can merge.

zner0L:
Ok, I finally checked and rebased this. This seems to be working fine. I am happy with merging if the tests come around. Do you wanna do the honors, @baltpeter?

baltpeter (approved):
Thank you!

PR #565, opened 2021-04-11
Email: Only generate reference-fragment if reference is set

mal-tee:


baltpeter:
Thanks for the PR but how did you manage to end up with no reference set? :o That really shouldn't be possible. If it is, that is the real bug we need to fix, as this definitely isn't the only part of the code depending on the reference being set.

mal-tee:
I thought you were aware of this bug :D

1. Make a request.
2. Go to my-requests and click on either admonition or complaint
3. Send the generated request as email --> you will get "(My reference: )" as subject

I am aware that this is only a hotfix, feel free to fix the RG(B), but this at least removes the useless text in the brackets for the time being :D

mal-tee:
Fun fact: The reference properly shows up in the generated request :)

baltpeter:
I have opened #570 to fix this properly.

PR #580, opened 2021-04-28
Better link references

baltpeter:
This PR fixes a surprising amount of our i18n-related link problems.


baltpeter:
This is now ready for review. Since the PR touches a large number of files and is a little "dangerous", it would be great if you could both do a review, @zner0L and @mal-tee.

content/de/act/ada-health/index.md
```diff
@@ -14,7 +14,7 @@ Die App *Ada Health* stellt sich auf ihrer [Webseite](https://ada.com/de/) als 
...
 
-Doch leider stellt sich heraus, dass ein genauerer Blick hinter die Kulissen nötig ist. Die Informationen, die über die App gesammelt und an den Anbieter geschickt werden, sind höchst sensible Daten. Nicht umsonst stellt die **DSGVO** (**Datenschutz-Grundverordnung**) Gesundheitsdaten unter [besonderen Schutz]({{< ref "gdpr-cheat-sheet" >}}#besondere-kategorien-personenbezogener-daten). Hier ist es umso wichtiger, dass die Verarbeitung sorgfältig und nur im nötigen Rahmen geschieht, um Verbraucher_innen zu schützen und unnötige Gefahren zu vermeiden.
+Doch leider stellt sich heraus, dass ein genauerer Blick hinter die Kulissen nötig ist. Die Informationen, die über die App gesammelt und an den Anbieter geschickt werden, sind höchst sensible Daten. Nicht umsonst stellt die **DSGVO** (**Datenschutz-Grundverordnung**) Gesundheitsdaten unter {{< link slug="gdpr-cheat-sheet#besondere-kategorien-personenbezogener-daten" text="besonderen Schutz" >}} Hier ist es umso wichtiger, dass die Verarbeitung sorgfältig und nur im nötigen Rahmen geschieht, um Verbraucher_innen zu schützen und unnötige Gefahren zu vermeiden.
```
mal-tee, line 17: Missing point.

mal-tee (changes requested):
Wow, that's a lot! :o Thank you!

mal-tee (approved):

baltpeter:
The new links added in #595 will need to be updated as well.

baltpeter:
I have made the necessary changes. @mal-tee, if you could review them, I think we change merge this without @zner0L's review.

layouts/partials/functions/link.html
```diff
@@ -0,0 +1,48 @@
...
+{{ if not .translate }}{{ $translate = false }}{{ end }}
+
+{{ $page := (site.GetPage .slug) }}
+{{ $text := (cond $translate (T .text .) .text) }}
+{{ if not $page }}
+    <!-- If the page doesn't exist in our current language, we fall back to English and apply a suffix to the text. -->
+    <!-- This depends on English being at index 1, which is highly unlikely to ever change. The alternative would have
+         been an ugly loop. -->
+    {{ $page = ((index site.Sites 1).GetPage .slug) }}
+    {{ $text = (printf "%s (EN)" $text) }}
```
zner0L, line 44: I am thinking this suffix should be wrapped in a `<span>` so that we can style it differently.

zner0L (approved):
Thank you, this looks really good.

layouts/partials/functions/link.html
baltpeter: Hm, the problem is that we can't do that clean here. We could only put raw HTML into the string and would then have to remember to `markdownify` that downstream for all users.

Do you have a specific styling in mind? Otherwise I would propose to postpone that change until someone actually wants to style it.

zner0L: I am fine with postponing, I just wanted to note this.

zner0L (approved):

PR #584, opened 2021-05-02
test language selection warning

mal-tee:


baltpeter:
Thanks! As is, the test is unfortunately not too useful. I know that's due to the same origin restrictions in Cypress. The typical workaround would be to simply test the `href` of the button instead of actually navigating there.

In this case, the redirect happens in the `onclick` handler, so it isn't quite as easy. However, all the `onclick` handler does is change `window.location`:

https://github.com/datenanfragen/website/blob/5afe9f908dcce545f825b8dcf904d5f456ce7aa7/src/Components/I18nWidget.js#L36

Thus, there is really no reason to not just replace it with a link. That would also have the added benefit of being more semantically correct and be better for a11y.

mal-tee:
I agree. Didn't know that replacing the handler in the Modal was that trivial. :D
I extended the test accordingly, but this should now be merged after #564 due to `baseUrl_DE`.

zner0L (approved):
Thanks, this can be merged right after #564.

PR #590, opened 2021-05-07
Company list: Use anchors as search suggestions (fixes #544)

### LINKED ISSUE #544: Search: Enable open in background
You can't open the results of the search in the background. We should consider converting the results to anchors where it makes sense, e.g. in the company database search.

grafik

mal-tee:


baltpeter:
Not sure if we want to do that but here is an interesting, slightly different approach to a large link that I have just seen: https://youtu.be/qKRq-tFixWE?t=2227

zner0L (changes requested):
Thanks! This looks good to me and can be merged, as soon as the changes I mentioned are adressed.

zner0L:
I forgot to mention that I also really like the way @baltpeter suggested for the links. That seems like a mich better UX for screen reader users. If you want to, implement it as well. Otherwise we could make a Hacktoberfest issue to change it.

mal-tee:
Turn it into an issue :+1:

src/Components/SearchBar.js
```diff
@@ -83,12 +83,15 @@ if (Privacy.isAllowed(PRIVACY_ACTIONS.SEARCH)) {
                         suggestion:
                             this.props.suggestion_template ||
                             function (suggestion) {
```
zner0L, line 85: Since you use `function` instead of arrow functions `() => {}`, the context is not correctly bound to the function (I know, I didn't expect there to be any semantic difference there either, when I first encountered this. That's JavaScript for you). That is why the test fails. Please use an arrow function to fix this.

zner0L (changes requested):
After you fixed the error in the test, I can merge this. If you also rebase before you upload, then I only need to do the rest of the merge?

src/Components/SearchBar.js
```diff
@@ -82,13 +82,16 @@ if (Privacy.isAllowed(PRIVACY_ACTIONS.SEARCH)) {
...
+                                const d = suggestion.document;
 
-                                let name_hs = suggestion.highlights.filter((a) => a.field === 'name');
-                                let runs_hs = suggestion.highlights.filter((a) => a.field === 'runs');
+                                const name_hs = suggestion.highlights.filter((a) => a.field === 'name');
+                                const runs_hs = suggestion.highlights.filter((a) => a.field === 'runs');
 
                                 return (
+                                    (this.props.anchorize
+                                        ? `<a class="no-link-decoration" href="${BASE_URL}/company/${d.slug}>`
```
zner0L, line 93:
```suggestion
                                        ? `<a class="no-link-decoration" href="${BASE_URL}/company/${d.slug}">`
```

zner0L (changes requested):

zner0L (approved):

PR #593, opened 2021-05-16
`postcss-url`: Don't inline every image

mal-tee:
* The config was broken before.
* This is especially useful for the bg-wave on the homepage.
  Now it's in a separate file, which saves us some KB in the main.css.


postcss.config.js
```diff
@@ -13,19 +18,15 @@ module.exports = (ctx) => {
             require('cssnano')({}),
             require('postcss-url')([
                 {
-                    filter: /\.svg/,
+                    filter: /\.(svg|png|jpg|gif|eot|ttf|woff|woff2)$/,
                     basePath: path.resolve(__dirname, 'assets/'),
-                    url: 'inline',
```
zner0L, line 18: We still want some assets to be inlined, don't we? I don't know what `postcss-url` normally does in this case, are files that are considered "small enough" inlined?

zner0L (changes requested):
Something is not working right. The `wave-bg.svg` gives me a 404 for the [Deploy Preview](https://deploy-preview-593--datenanfragen.netlify.app/).

zner0L:
@mal-tee Do you mind taking another look at this?

mal-tee:
Locally it inlines the asterisk for required form fields and the bg-wave is loaded successfully from a file. I don't know why it doesn't work in the CI.

If someone wants to pick up the pieces: Feel free to do so. I won't for the foreseeable future. Sorry.

PR #594, opened 2021-05-20
More speed optimizations: Images again, fonts CSS

baltpeter:


baltpeter:
This is now ready for review.

mal-tee:
Sidenote: I think we could lazy load the itz-logo as well?

https://github.com/datenanfragen/website/blob/b_make-google-happy/layouts/_default/baseof.html#L175

baltpeter:
> Side note: I think we could lazy load the itz-logo as well?

Sure.

mal-tee (dismissed):
Wow, very nice! Thank you! What do you think: Is making tests for the "image pipeline" worth the effort?

baltpeter:
> What do you think: Is making tests for the "image pipeline" is worth the effort?

What would you want to test in particular?

mal-tee:
Alt texts, correct height/width-ratio, that an image is shown in the first place, not the scaling or format.

cypress/integration/components/hugo-images.spec.js
```diff
@@ -0,0 +1,66 @@
...
+            ]) {
+                const source = win.document.querySelector(`picture source:nth-of-type(${num})`);
+                expect(source.srcset).to.include(format);
+                expect(source.srcset).to.not.include(format === '.webp' ? '.jpg' : '.webp');
+            }
+        });
+    });
+
+    it('Featured images are not lazy-loaded', () => {
+        cy.get('picture img').should('not.have.attr', 'loading');
```
mal-tee, line 60:
```suggestion
        cy.get('picture img').first().should('not.have.attr', 'loading');
```

Seems cleaner to me.

mal-tee (approved):
Thank you!

PR #601, opened 2021-05-23
Fixes #182: Don't remember clearing the signature in the generator  

### LINKED ISSUE #182: Bug: Saved signature is overridden on reset
In the generator, if the signature for the request is reset the blank signature is also save as the saved signature. This behavior is counter intuitive IMO. The expected behavior would be for the current request to have a blank signature field and the saved signature should be kept, if the reset button is clicked in the generator, but the saved signature is overridden, if the reset button is clicked on the saved id data page.

baltpeter:
This also fixes quite a severe bug with the `SignatureInput` (see commit message).

I haven't implemented any tests in this PR as I'm not even sure if that's possible with Cypress.


mal-tee (approved):
Thank you!

PR #602, opened 2021-05-23
Closes #504: Fix cut-off 'required' asterisk

### LINKED ISSUE #504: 'Required' asterisk in inputs is cut off
The star is cut off.

_Originally posted by @mal-tee in https://github.com/datenanfragen/backend/pull/23#discussion_r520627733_

baltpeter:
The view box of the SVG was not big enough.

Also change font from FiraGO to Arial as, at least for me, FiraGO
isn't available in the SVG and instead replaced by a default font
that looks quite ugly here.
Arial should be available on just about every system and also ensures
that the asterisk has a consistent height and isn't cut off (again)
depending on the default font per system.

### DIFF
--- a/assets/styles/elements.scss
+++ b/assets/styles/elements.scss
@@ -92,8 +92,8 @@ canvas {
 // mean, good on you for adhering to the spec. But still. Come on.
 // Anyway, that is why we are URL-encoding the `#` here (inspired by: https://codepen.io/kevinweber/pen/dXWoRw).
 $required-asterisk-svg: str-replace(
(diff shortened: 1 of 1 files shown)


mal-tee (changes requested):
Thank you, but now the asterisk looks cut-off on the top. I checked it in FF, Chrome and Edge.

image

mal-tee:
It's also not vertically aligned I think? My guess: Both is due to the fact that the asterisk char is in superscript?

baltpeter:
Interesting. Renders fine for me.

image

baltpeter:
Let's try a hard-coded path, then.

mal-tee (approved):
Thank you, that works! They are a bit large now, but I guess that's just something to get used to.

PR #603, opened 2021-05-23
DonationWidget: Don't print bank transfer QR codes (fixes #494)

### LINKED ISSUE #494: Donate: QR code titles are printed but not the codes themself
Why print the QR codes at all?

baltpeter:


mal-tee (approved):
Thank you very much!

PR #606, opened 2021-05-29
Fix 'no-empty-date-in-generator' test failing in production

baltpeter:
`window.accessLocalForageStore()` comes from `test-interface.js`,
which is not available on the production site. Thus, we need to skip
this test there.


mal-tee (approved):
Thank you!

PR #613, opened 2021-06-05
Error-handler: check if browser is online, fixes #528

### LINKED ISSUE #528: Error Handling: Check the internet connection
Before showing the big error modal, we could check if the exception in question is connection related and if the client has no active internet connection.

mal-tee:


src/error-handler.js
```diff
@@ -1,3 +1,5 @@
+import FlashMessage, { flash } from 'Components/FlashMessage';
```
baltpeter, line 1: Not too happy about this to be honest. We are deliberately not including anything in the error handler to make sure that it stays as small as possible and also avoid as many potential points of failure as possible.

And as you said, the UX doesn't seem to be great anyway. Maybe we can think of another way to implement this?

src/error-handler.js
```diff
@@ -67,9 +67,46 @@ function primitiveErrorModal(enduser_message, github_issue_url, mailto_url) {
     document.querySelector('#error-modal .close-button').onclick = dismiss;
 }
 
+const flash_message_id = 'flash-message-error-online';
+const button_id = 'button-flash-message-error-online';
 try {
     const handler = (event) => {
         try {
+            if (event.error.message === 'Network Error') {
```
baltpeter, line 75: Most things I have tried (submitting comments, donations, loading templates in the generator) produce another error and are thus not caught by the current code.

In Chrome, the message seems to be "Uncaught TypeError: Failed to fetch" and in Firefox, it seems to be "TypeError: NetworkError when attempting to fetch resource.".

baltpeter (changes requested):

baltpeter (approved):
Thank you!

PR #614, opened 2021-06-05
test bank transfer donations

mal-tee:


cypress/integration/use-cases/donate.spec.js
```diff
@@ -0,0 +1,31 @@
...
+        cy.get('#bank-transfer-data-table > tr:nth-child(2) > td:nth-child(2)').should(
+            'have.text',
+            'DE42 8306 5408 0104 0851 40'
+        );
+        cy.get('#bank-transfer-data-table > tr:nth-child(1) > td:nth-child(2)').should(
+            'contain.text',
+            'Datenanfragen.de e. V.'
+        );
+        cy.contains('Donation verification').click();
+        cy.url().should('include', '/thanks');
```
baltpeter, line 21: We should also check that the donation reference is included correctly. Both here and in the `mailto` link.

cypress/integration/use-cases/donate.spec.js
```diff
@@ -0,0 +1,43 @@
+describe('Using the donate function', () => {
+    it('lets users donate via bank transfer', () => {
+        cy.visit('/donate');
+        cy.contains('10 €').click();
+        cy.get('#donation-widget-amount').should('have.value', 10);
```
baltpeter, line 5: Let's also test that the text to the right ("You will be billed 10.00 € once.") contains the correct value.

baltpeter (commented):
One last request. :) After that, we're good to merge.

cypress/integration/use-cases/donate.spec.js
```diff
@@ -0,0 +1,46 @@
...
+        cy.get('#donation-widget-amount-section > div:nth-child(3) > span:nth-child(1)')
+            .invoke('text')
+            .should('match', /\b10.00\b/);
```
baltpeter, line 8:
```suggestion
            .should('match', /\b10\.00\b/);
```

cypress/integration/use-cases/donate.spec.js
```diff
@@ -0,0 +1,46 @@
...
+        cy.contains('Bank transfer').click();
+        cy.contains('To the next step').click();
+        cy.contains('Payment information');
+        cy.get('#bank-transfer-data-table > tr:nth-child(6) > td:nth-child(2)')
+            .invoke('text')
+            .should('match', /^10.00/);
```
baltpeter, line 14:
```suggestion
            .should('match', /^10\.00/);
```

baltpeter (changes requested):

Issue #582, opened 2021-04-29
Discoverability of the other language versions


mal-tee:
> Also, please check the possibility of automatically opening the "localized" version of the website (depending upon the browser's language preferences/settings) when a user enters any of the base URL names in the browsers URL address field (no matter which "translated" site name may be specified - datarequests, datenanfragen, osobnipodaci, demandetesdonnees, pedidodedados, solicituddedatos, or any future translated name). I'm sure this will be difficult ... ;-(

That's actually something we decided against. We find automatic language redirects a lot more annoying than helpful. But I agree that the discoverability of the other language versions could be better. If you have an idea (design-wise, the implementation should actually not be too difficult) for an unobtrusive way of introducing those to users visiting the "wrong" language version, we can gladly discuss that in a new issue.

_Originally posted by @baltpeter in https://github.com/datenanfragen/website/issues/577#issuecomment-829262293_

mal-tee:
I agree, we should improve the discoverability.

How about something like a hint in the top of the screen indicating that the page is available in a different language?

baltpeter:
> How about something like a hint in the top of the screen indicating that the page is available in a different language?

Do you want to create a mockup or code a quick demo of what that would look like?

mal-tee:
I am not a designer, but sure, here is my take:

![grafik](https://user-images.githubusercontent.com/45886910/116731812-7aa4e380-a9ea-11eb-82de-fe3312147326.png)

Maybe the little box should also get a different background-color or a little flag corresponding to the language. I am not sure about the flag, on the one hand it is widely known as entry point for configuring languages, on the other hand it's not really meant to represent a language in the first place.

---

I was thinking of only showing this if the current language mismatches the browsers language setting. If you dismiss it once its gone for good.

milotype:
> I was thinking of only showing this if the current language mismatches the browsers language setting. If you dismiss it once its gone for good.

I agree ... but I'd suggest to use a "popup".
I agree ... don't use country flags.

**My suggestion**

Text for the popup:
"Change the website's language to `preferred-language` or `select` your preferred language."

- `preferred-language` should be substituted with the "translated" string for the language (Hrvatski or Deutsch, etc.) corresponding to the user's web-browser preferences. If selected, the corresponding website opens. Popup automatically closes.
[...]

PR #615, opened 2021-06-06
Notify user of other languages available

### LINKED ISSUE #270: Follow ISO language codes recommandations
Languages code for German, English and French are currently formatted like this: "de-de", "en-us" and "fr-fr". I think it should be respectively "de-DE", "en-US" and "fr-FR" as the first two letters refers to the language code and the two others to the country code. Generally the country code should be in uppercase.

I don't think the country code is necessary for countries where the language comes from. For example "fr" stands for French (France) in any case even if not written "fr-FR". However "fr-CA" stands for French (Canada). Same example with "pt" for Portuguese (Portugal) and "pt-BR" for Portuguese (Brazil). But it can be more consistent to always write the full language code and so it can stay like this!

davenewham:
A working solution for #582 
Also closes #270 

I would prefer the text to be translated in the target language - but I can't seem to get the t_r function working.https://github.com/datenanfragen/website/blob/ab306ea794af818e051489860fcc805f50561bd4/src/Utility/i18n.js#L7-L10 . However, it seems like the ideal solution to presenting the text in the langauge that we are aiming for

Anyway an example on the German website recommending other langauges:

image
image

It currently only works the first time the user visits the webpage - before the lang cookie is set. There are no subsequent checks. The unit tests will be fun to write for this! 

Looking for any feedback/recommendations :)


src/Components/I18nWidget.js
```diff
@@ -21,15 +21,17 @@ export default class I18nWidget extends Component {
...
         if (I18nWidget.alreadyShowingModal || event.target.value === LOCALE) return;
-
+        
         let modal;
         const dismiss = () => {
             render('', document.body, modal);
             I18nWidget.alreadyShowingModal = false;
             event.target.value = LOCALE;
         };
+        let selected_lang = event.target.value; //e.g hr
```
davenewham, line 31: no fallback if event.target.value is null

cypress/integration/use-cases/language-selection.spec.js
```diff
@@ -5,9 +5,9 @@ describe('switching languages', () => {
         cy.get('footer .i18n-widget-language select').select('German (Deutsch)', {
             force: true,
         });
-        cy.contains('Changing the language will redirect you to another domain.');
+        cy.contains('Wenn Du die Sprache wechelst, wirst Du zu einer anderen Domain weitergeleitet.');
```
davenewham, line 8: Need to add route for german -> english, also both text should appear, what if the user clicks the wrong language by accident?

src/i18n/en.json
```diff
@@ -524,12 +528,13 @@
...
         "language-hr": "Croatian",
-        "language-change-modal": "Changing the language will redirect you to another domain. Your settings and saved data from this domain will not carry over.<br>In particular, this means that things like your privacy settings and saved requests will not be available in the new language. You can still always go back to this language and view them here but you cannot share settings and saved data between different languages.",
-        "change-lang": "Change language",
-        "stay": "Stay here",
         "not-supported-prompt": "Your language or country is not supported yet?",
         "not-supported-cta": "Help us support it!"
     },
+    "interlingual": {
+        "recommend-en": "This site is also available in your langauge!",
+        "recommend-de": "Diese Seite ist auch in Ihrer Sprache verfügbar!"
```
mal-tee, line 536:
```suggestion
        "recommend-de": "Diese Seite ist auch in Deiner Sprache verfügbar!"
```

We use the informal "you" :D

davenewham: True - I think this however should now be moved to each of the language files in the requests section similar to how I'm doing it for the i18n element

davenewham: Resolved in a0aa8fb

src/Components/I18nWidget.js
```diff
@@ -28,8 +28,10 @@ export default class I18nWidget extends Component {
             I18nWidget.alreadyShowingModal = false;
             event.target.value = LOCALE;
         };
+        let selected_lang = event.target.value || LOCALE; //e.g hr
```
baltpeter, line 31:
```suggestion
        const selected_lang = event.target.value || LOCALE;
```

src/general.js
```diff
@@ -42,6 +44,22 @@ if (comments_div) {
...
+ * but the browser is configured in German, lets notify them to change the lang
+ * @param  {string} preferred_language - bcp47 substring of language. e.g en-US becomes en
+ * @param  {string} website_language - bcp47 substring of current website language eg en-US becomes en
+ */
+function notifyOtherLanguages(preferred_language, website_language) {
+    if (!preferred_language || !website_language) return;
+    let recommend_language = t_r('recommend-language', preferred_language || LOCALE);
+    flash(
+        <FlashMessage type="info" duration={15000}>
+            {recommend_language} {<I18nWidget minimal={true} showLanguageOnly={true} />}
```
baltpeter, line 58: I'm pretty sure you don't need to put the JSX element in curly braces.

src/general.js
```diff
@@ -42,6 +44,22 @@ if (comments_div) {
...
+ * Notify the user that other languages may exist from their current browser. Ie if someone visits the English site,
+ * but the browser is configured in German, lets notify them to change the lang
+ * @param  {string} preferred_language - bcp47 substring of language. e.g en-US becomes en
+ * @param  {string} website_language - bcp47 substring of current website language eg en-US becomes en
+ */
+function notifyOtherLanguages(preferred_language, website_language) {
+    if (!preferred_language || !website_language) return;
+    let recommend_language = t_r('recommend-language', preferred_language || LOCALE);
+    flash(
+        <FlashMessage type="info" duration={15000}>
```
baltpeter, line 57: I think we can have the duration a lot longer (maybe even `INT_MAX`?) as this is only shown once. I was pretty annoyed in my testing that it disappeared so quickly and that will probably be a lot more annoying for a user who doesn't already know what the notification is about.

src/general.js
```diff
@@ -58,6 +76,17 @@ function guessUserCountry() {
     const bcp47_regex = /^(?:(en-gb-oed|i-ami|i-bnn|i-default|i-enochian|i-hak|i-klingon|i-lux|i-mingo|i-navajo|i-pwn|i-tao|i-tay|i-tsu|sgn-be-fr|sgn-be-nl|sgn-ch-de)|(art-lojban|cel-gaulish|no-bok|no-nyn|zh-guoyu|zh-hakka|zh-min|zh-min-nan|zh-xiang))$|^((?:[a-z]{2,3}(?:(?:-[a-z]{3}){1,3})?)|[a-z]{4}|[a-z]{5,8})(?:-([a-z]{4}))?(?:-([a-z]{2}|\d{3}))?((?:-(?:[\da-z]{5,8}|\d[\da-z]{3}))*)?((?:-[\da-wyz](?:-[\da-z]{2,8})+)*)?(-x(?:-[\da-z]{1,8})+)?$|^(x(?:-[\da-z]{1,8})+)$/i;
     const bcp47_country = (bcp47_regex.exec(navigator_lang)[5] || '').toLowerCase();
 
+    const bcp47_preferred_language = (bcp47_regex.exec(navigator_lang)[3] || '').toLowerCase();
+    const bcp47_website_language = (bcp47_regex.exec(document.documentElement.lang)[3] || '').toLowerCase();
+
+    // Suggest user other language
+    if (bcp47_preferred_language != bcp47_website_language) {
```
baltpeter, line 83:
```suggestion
    if (bcp47_preferred_language !== bcp47_website_language) {
```

src/general.js
```diff
@@ -42,6 +44,22 @@ if (comments_div) {
     );
 }
 
+/**
+ * Notify the user that other languages may exist from their current browser. Ie if someone visits the English site,
+ * but the browser is configured in German, lets notify them to change the lang
+ * @param  {string} preferred_language - bcp47 substring of language. e.g en-US becomes en
+ * @param  {string} website_language - bcp47 substring of current website language eg en-US becomes en
```
baltpeter, line 51:
```suggestion
 * Notify the user that other languages may exist from their current browser. I.e. if someone visits the English site,
 * but the browser is configured in German, let's notify them to change the lang.
 * @param  {string} preferred_language - bcp47 substring of target language, e.g. `en-US` becomes `en`
 * @param  {string} website_language - bcp47 substring of current website language, e.g. `en-US` becomes `en`
```

baltpeter (changes requested):
Thanks, @davenewham. A few minor comments below. Other than the failing test, there's no big blockers anymore.

src/general.js
davenewham: Tried this with Number.MAX_SAFE_INTEGER  and Number.MAX_VALUE, both didn't work

davenewham: Updated it to now 90 seconds

baltpeter: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#maximum_delay_value), the max value for `setTimeout()` (which `FlashMessage` is using to dismiss itself) is `2147483647` which is a lot smaller than `Number.MAX_SAFE_INTEGER` (`9007199254740991`) but is actually C's `INT_MAX`, so I wasn't completely off. :D 

Please either hardcode the duration as `2^31-1` or change `FlashMessage` to interpret a duration of `-1` as "never dismiss automatically".

cypress/integration/use-cases/language-selection.spec.js
```diff
@@ -5,10 +5,35 @@ describe('switching languages', () => {
...
+        cy.visit('/', {
+            onBeforeLoad(win) {
+                Object.defineProperty(win.navigator, 'language', {
+                    value: ['en-GB'],
+                });
+            },
+        });
+        cy.get('#flash-messages').should('not.contain', 'Diese Seite ist auch in Deiner Sprache verfügbar!');
+    });
+});
```
baltpeter, line 39: Why'd you remove the test with the `country` cookie set? That seems important. :D

davenewham: Oops, my bad. Added back in #2fb7811

src/i18n/en.json
```diff
@@ -511,7 +516,6 @@
     },
     "i18n-widget": {
         "country": "Country",
-        "language": "Language",
```
baltpeter, line 514: Oh sorry, one last problem I missed:

This string also needs to stay here (in all languages). Otherwise, the heading in the "regular" I18nWidet is missing:

image

baltpeter: I'll just fix that myself, so we can finally get this PR merged. :D

davenewham: Oops, that's my bad. Thanks

baltpeter:
```diff
diff --git a/src/i18n/de.json b/src/i18n/de.json
--- src/i18n/de.json
+++ src/i18n/de.json
@@ -515,8 +515,9 @@
         "overlay-allow": "Erlauben"
     },
     "i18n-widget": {
         "country": "Land",
+        "language": "Sprache",
         "language-desc-de": "Deutsch",
         "language-desc-en": "Englisch (English)",
         "language-desc-fr": "Französisch (Français)",
         "language-desc-es": "Spanisch (Español)",
diff --git a/src/i18n/en.json b/src/i18n/en.json
--- src/i18n/en.json
+++ src/i18n/en.json
@@ -515,8 +515,9 @@
         "overlay-allow": "Allow"
     },
     "i18n-widget": {
         "country": "Country",
+        "language": "Language",
         "language-desc-de": "German (Deutsch)",
         "language-desc-en": "English",
         "language-desc-fr": "French (Français)",
         "language-desc-es": "Spanish (Español)",
diff --git a/src/i18n/es.json b/src/i18n/es.json
--- src/i18n/es.json
+++ src/i18n/es.json
@@ -730,8 +730,9 @@
         "language-desc-fr": "Francés (Français)",
         "language-desc-en": "Inglés (English)",
         "language-desc-de": "Alemán (Deutsch)",
         "language-desc-es": "Español",
+        "language": "Idioma",
         "country": "País"
     },
     "signature": {
         "overlay-allow": "Permitir",
diff --git a/src/i18n/fr.json b/src/i18n/fr.json
--- src/i18n/fr.json
+++ src/i18n/fr.json
@@ -308,8 +308,9 @@
         "overlay-allow": "Autoriser"
     },
     "i18n-widget": {
         "country": "Pays",
+        "language": "Langue",
         "language-desc-de": "Allemand (Deutsch)",
         "language-desc-en": "Anglais (English)",
         "language-desc-fr": "Français",
         "language-desc-es": "Espagnol (Español)",
diff --git a/src/i18n/hr.json b/src/i18n/hr.json
--- src/i18n/hr.json
+++ src/i18n/hr.json
@@ -841,7 +841,8 @@
         "language-desc-fr": "Francuski (Français)",
         "language-desc-en": "Engleski (English)",
         "language-desc-de": "Njemački (Deutsch)",
         "language-desc-hr": "Hrvatski",
+        "language": "Jezik",
         "country": "Zemlja"
     }
 }
diff --git a/src/i18n/pt.json b/src/i18n/pt.json
--- src/i18n/pt.json
+++ src/i18n/pt.json
@@ -474,8 +474,9 @@
         "overlay-allow": "Autorizar"
     },
     "i18n-widget": {
         "country": "País",
+        "language": "Idioma",
         "language-desc-de": "Alemão (Deutsch)",
         "language-desc-en": "Inglês (English)",
         "language-desc-fr": "Francês (Français)",
         "language-desc-es": "Espanhol (Español)",

```

baltpeter (approved):
Thanks again.

PR #623, opened 2021-07-04
Update sample letters according to datenanfragen/data#1167 (explicitly request copy)

baltpeter:
see datenanfragen/data#1167


mal-tee (approved):
Thanks!

PR #629, opened 2021-07-25
Display what's left of a donation after fees, fixes #386

### LINKED ISSUE #386: DonationWidget: Show the amount that we will receive after fees
We offer multiple payment methods in our donation widget that have different fees. Users may want to choose the payment method with the smallest fees for a more effective use of their money.

For reference, this is how [FragDenStaat](https://fragdenstaat.de/spenden/) does this:

image

mal-tee:
- I will need some real values for the fees :) 

- Is there a specific reason we didn't use JS' `Intl.NumberFormat` in `renderMoney` instead of `{{amount}}&nbsp;€` in our templates?


src/Components/DonationWidget.js
```diff
@@ -120,7 +132,29 @@ export default class DonationWidget extends Component {
                                     onChange={(e) => {
                                         this.setState({ payment_method: e.target.value });
                                     }}
-                                    label={<Text id={payment_method} />}
+                                    label={
+                                        <div>
+                                            <Text id={payment_method} />
+                                            {PAYMENT_NETTO[payment_method] && (
+                                                <p style="text-align:right; margin: 0px; margin-top: -10px;">
```
baltpeter, line 139:
```suggestion
                                                <p style="text-align: right; margin: -10px 0 0 0;">
```

src/i18n/de.json
```diff
@@ -643,6 +643,7 @@
         "bank-transfer-amount": "Betrag",
         "bank-transfer-qrcodes": "QR-Codes für die Überweisung",
         "amount": "Dir werden <em>einmalig</em> {{amount}}&nbsp;€ abgebucht.",
+        "amount-after-fees": "Wir kriegen davon {{amount}}&nbsp;€",
```
baltpeter, line 646: I would like to make it more clear that this is due to fees as that may not be apparent to everyone. Maybe like this?

```suggestion
        "amount-after-fees": "Nach Gebühren kriegen wir {{amount}}&nbsp;€.",
```

src/i18n/en.json
```diff
@@ -643,6 +643,7 @@
         "bank-transfer-amount": "Amount",
         "bank-transfer-qrcodes": "QR codes for the transfer",
         "amount": "You will be billed {{amount}}&nbsp;€ <em>once</em>.",
+        "amount-after-fees": "We'll get {{amount}}&nbsp;€",
```
baltpeter, line 646:
```suggestion
        "amount-after-fees": "After fees, we’ll get {{amount}}&nbsp;€.",
```

baltpeter:
>     * I will need some real values for the fees :)

* Bank transfers: free
* [CoinGate](https://coingate.com/pricing): 1%
* [PayPal](https://www.paypal.com/de/webapps/mpp/charity): 1.5% + 0.35€
* [Mollie](https://www.mollie.com/en/pricing): Depends on the method, unfortunately. :/

>     * Is there a specific reason we didn't use JS' `Intl.NumberFormat` in `renderMoney` instead of `{{amount}}&nbsp;€` in our templates?

I don't think so, feel free to change.

src/Components/DonationWidget.js
```diff
@@ -12,6 +12,18 @@ const DONATIONS_API = 'https://backend.datenanfragen.de/donation';
 const SUGGESTED_AMOUNTS = [5, 10, 15, 25, 50, 75, 100, 150, 200, 250];
 const PAYMENT_METHODS = ['bank-transfer', /*'creditcard',*/ 'cryptocurrency', 'paypal', 'mollie'];
 
+const linear_func = (fee_percent, fee_fixed, x) => {
+    return x - fee_fixed - x * fee_percent;
+};
```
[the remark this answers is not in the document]
baltpeter: The name isn't too descriptive. Maybe something like `amountAfterFees()` instead?

src/Components/DonationWidget.js
baltpeter: This unfortunately doesn't work well on mobile.

src/Components/DonationWidget.js
```diff
@@ -120,7 +132,29 @@ export default class DonationWidget extends Component {
...
+                                                    <small>
+                                                        <MarkupText
+                                                            id="amount-after-fees"
+                                                            fields={{
+                                                                amount: renderMoney(
+                                                                    Math.floor(
+                                                                        PAYMENT_NETTO[payment_method](
+                                                                            this.state.amount
+                                                                        ) * 100
+                                                                    ) / 100
```
baltpeter, line 149: The rounding here is not necessary.

baltpeter (changes requested):

zner0L:
The CoinGate fees also depend on the current Cryptocoin price volatility, in case you want to factor that in ;P

baltpeter (approved):
Thank you!

PR #631, opened 2021-07-27
Update limit for simplified donation receipt to 300 €

baltpeter:
The limit has been raised at the start of 2021.


mal-tee (approved):
Thanks!

PR #634, opened 2021-08-07
Resolves #609: Added widths and heights to some SVGs

### LINKED ISSUE #609: Set explict `width` and `height` for SVGs in blog posts
As Hugo's `imageConfig` unfortunately [doesn't support SVG](https://github.com/gohugoio/hugo/issues/3700) this will have to be done manually.

An example of where this problem occurs is the sample letters overview page. Here's what [PageSpeed](https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fwww.datenanfragen.de%2Fsample-letters%2F) says about that:

image

thewindsofwinter:
Hello,

I've updated the widths and heights of SVGs, which should at least partially fix issue #609. There appear to be a few kinds of changes in the files:

 - The changes that I actually meant to make (adding widths and heights)
 - Some files may have been CRLF (Windows line endings) before and I turned them into LF while trying to standardize all the line endings: I think this should be fine unless some files must have CRLF line endings
 - I also was able somehow to run the hooks/linting before committing so some minor format changes seem to have occurred.

Notes:

 - I'm not the greatest with Git, so I haven't bothered to squash and rebase before merging, please let me know if I need to do things like that. (Was just worried it would wipe out my repo bc I'm bad with Git).
 - I haven't had the time to look through further for more SVGs, but I thought I'd submit this PR anyways, because school is starting up and I doubt I'll have that much time in the future.

Thanks, and I eagerly await your review.
~Andy

### DIFF
--- a/content/de/sample-letters.md
+++ b/content/de/sample-letters.md
@@ -13,7 +13,7 @@ So wird das Stellen von Anfragen einfach: Du musst nur noch die entsprechenden D
 Die Mustertexte unterscheiden sich natürlich bei den unterschiedlichen Anfragetypen. Wir unterstützen die wichtigsten davon.
 
 <article class="list-article icon-list-article">
-    <div class="col25 article-featured-image"><a href="/blog/musterbrief-dsgvo-anfrage-auskunft/"><img class="image" src="/card-icons/view.svg" alt="Antrag auf Auskunft nach Art. 15 DSGVO"></a></div>
+    <div class="col25 article-featured-image"><a href="/blog/musterbrief-dsgvo-anfrage-auskunft/"><img class="image" src="/card-icons/view.svg" alt="Antrag auf Auskunft nach Art. 15 DSGVO" width="252" height="150"></a></div>
     <div class="padded col75">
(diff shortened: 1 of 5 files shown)


baltpeter (approved):
Thanks for the PR, @thewindsofwinter!

I have left some comments below.

While the unrelated changes you mentioned aren't a problem (and that's even how it should have been done in the first place), I would prefer not to have them in this PR to keep a clean commit history. If you want, I can clean that up later once the other changes are done.

> I'm not the greatest with Git, so I haven't bothered to squash and rebase before merging, please let me know if I need to do things like that. (Was just worried it would wipe out my repo bc I'm bad with Git).

No worries, unless there's any conflicts, GitHub does that automatically for us when merging.

baltpeter (changes requested):
Sorry, clicked the wrong button. :D

thewindsofwinter:
Sounds good, I will change ASAP! Thanks for the quick review, I'm still pretty new to this so hopefully I should be able to get better

thewindsofwinter:
Just fixed the things you requested, looks like there are a couple conflicts that need to be resolved, I will see what's wrong

PR #636, opened 2021-08-09
Fix overflowing links not breaking in Chrome

baltpeter:
In the course of our ongoing mission to appease Google, I have noticed that they have been complaining about mobile usability issues on the following pages:

* /company/einwohnermeldeamt-bs/
* /blog/ausweispflichten/

The problems actually don't occur in Firefox as we have set `overflow-wrap: break-word;`. But that doesn't seem to work in Chrome, at least for URLs:

image

image

So, instead we now set `overflow-wrap: anywhere;` for links, which still breaks them nicely in Firefox:

image

image

And at least breaks them at all in Chrome:

image

image

### DIFF
--- a/assets/styles/framework.scss
+++ b/assets/styles/framework.scss
@@ -196,6 +196,8 @@ a:focus {
 }
 
 main a {
+    overflow-wrap: anywhere;
+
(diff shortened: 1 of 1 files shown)


zner0L (approved):
Thanks!

PR #441, opened 2020-10-11
refocus search bar after selecting company

sumirart:
fixes  #435 
thanks!

PR #645, opened 2021-08-17
Revert #441 to working solution

zner0L:
Since I left #441 lying around for almost a year, I can't critique it appropriately anymore. I took it upon myself to revert to the old way, because it is the only way to capture the focus, since we don't use preact properly (I would love someone to fix that!).
I also made capturing the focus optional, because this might happen unexpectedly in some case, which would be horrible for our users using screen readers.

### DIFF
--- a/src/Components/SearchBar.js
+++ b/src/Components/SearchBar.js
@@ -140,7 +140,7 @@ if (Privacy.isAllowed(PRIVACY_ACTIONS.SEARCH)) {
             if (typeof this.props.setupPlaceholderChange === 'function') {
                 this.props.setupPlaceholderChange(this.input_element);
             }
-            this.input_element.focus();
+            if (this.props.captureFocus) this.input_element.focus();
         }
 
         componentWillUnmount() {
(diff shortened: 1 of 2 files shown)


baltpeter:
Is this even necessary anymore? I am already getting the exact same behaviour in prod without this PR:

https://user-images.githubusercontent.com/4048809/129776136-44ebd9e0-3d61-4ae4-aaf4-d910ca8ae369.mp4

This happens both in the wizard (where it makes sense) and in the generator (where it doesn't make too much sense). Iirc, this was changed in the Preact X PR?

Annoyingly though, the text in the search bar is not cleared. We should definitely do that, at least for the wizard.

zner0L:
Interesting. I tried this, but I run into the same problem as before, that preact doesn't refresh this the way you would expect. As I said, there is a more fundamental issue here in that we should really use a preact/react autocomplete component instead of duct taping a different library (that is pretty heavy even) to an input. I think we should close this in favor of fixing the fundamental problem.

baltpeter:
We did have an issue for implementing our own autocomplete component in our old issue tracker (https://git.my-server.in/datenanfragen/website/-/issues/24) but never got around to that as there is a lot more work involved than I initially thought.

I wouldn't mind switching to a different (P)React-native solution. Not sure if there are any good ones available.

> I think we should close this in favor of fixing the fundamental problem.

Sure, fine with me.

zner0L:
I moved this into #649 and will close this now.

PR #666, opened 2021-09-13
Fixes #656: Ignore errors from extensions in error handler

### LINKED ISSUE #656: Error handler: Ignore errors from extensions
As browser extensions (can) run in the context of the website, they can also cause errors in the context which would be caught by our error handler. This happened in #652.

As we can usually not do anything about those errors, we should ignore them in the error handler.

baltpeter:


mal-tee (approved):
Thank you very much!

PR #667, opened 2021-09-13
Fixes #620: Don't include Hugo translations in JS bundle

### LINKED ISSUE #620: Why are we including the Hugo translations in the JS translations bundle?
I haven't thoroughly checked this but if memory serves me correctly, the Hugo translations should not be used anywhere from JS. However, we are currently including them in the JS translations bundle (and they make up a significant section of it!)

This is a lot of wasted space that we should definitely get rid of.

baltpeter:
I am _reasonably_ confident that this shouldn't cause any issues. I also couldn't find any when looking through the code and the 'rendered' site. Nonetheless, this should be reviewed carefully.

### DIFF
--- a/src/Utility/webpack-i18n-loader.js
+++ b/src/Utility/webpack-i18n-loader.js
@@ -26,6 +26,9 @@ module.exports = function (content) {
         this.emitFile(path.join('..', 'i18n', path.basename(this.resource)), JSON.stringify(hugo_data));
     }
 
+    // The JS translation files don't need to include the translations only used by Hugo (#620).
+    delete data.hugo;
+
     // Emit the translation files to be included in the HTML.
     this.emitFile(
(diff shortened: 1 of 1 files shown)


mal-tee (approved):
I poked around the deploy preview and didn't find anything. Tests pass. I grepped `website` for "hugo" and didn't find anything of interest. So I guess we are good to go?

You might want to update this comment as well:

https://github.com/datenanfragen/website/blob/3ea0cae51c2d13163a81bb2581ee9b89f07d8f6f/webpack.common.js#L86-L92

baltpeter:
Thanks!

> You might want to update this comment as well:

I think the comment is still correct. If anything, it makes more sense now. I don't think anyone would have expected us to unnecessarily include the Hugo translations in the JS ones based on that comment.

PR #669, opened 2021-09-14
Use MarkupText for 'add-dynamic-input-explanation' in generator

baltpeter:
The French translation uses a '&nbsp;' here which otherwise breaks.


mal-tee (approved):
Thank you!

PR #692, opened 2021-09-23
higher image quality, tweak webp hint

mal-tee:
- the tweaked hint gets rid of the weird lines in, e.g. [this image](https://www.datenanfragen.de/blog/jahresbericht-2020/jahresbericht-2020_hu089e1766bccfbf267935c78330f990a2_40902_230x0_resize_q75_h2_box_2.webp)
	- **UPDATE**: seems like that didn't do the trick 	
- the higher quality (default is 75) gets rid of some compression artifacts, but comes with a slight size increase 
  - 15% or 0.77 KB for my [test thumbnail](https://www.datenanfragen.de/blog/datenloeschung-bei-werbetreibenden/datenloeschung-werbetreibende_hu9826f4f2c44a3b0acf4413123c531f73_198477_230x0_resize_q75_h2_box.webp) in my local build
  - 22% or 5.94KB for the [same image in big](https://www.datenanfragen.de/blog/datenloeschung-bei-werbetreibenden/datenloeschung-werbetreibende_hu9826f4f2c44a3b0acf4413123c531f73_198477_784x0_resize_q75_h2_box.webp)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,8 +22,8 @@ jobs:
   build:
     executor: docker_hugo_node
     environment:
-        - HUGO_VERSION: '0.85.0'
-        - HUGO_BINARY: hugo_extended_0.85.0_linux-64bit
+        - HUGO_VERSION: '0.88.1'
+        - HUGO_BINARY: hugo_extended_0.88.1_linux-64bit
     steps:
       - run:
           command: |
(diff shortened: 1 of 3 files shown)


mal-tee:
"Fix transparency problem when converting 32-bit images to WebP 8f40f34c @bep #8729" 
- https://github.com/gohugoio/hugo/releases/tag/v0.86.0
- https://github.com/gohugoio/hugo/issues/8729

baltpeter:
I'm definitely on board for changing the hint an updating Hugo. The size increase for the thumbnails is also ok with me.

As for the regular images, that is a more significant increase. I just did a little non-scientific blind test trying to guess which of the versions is the "better" one and while I can see small differences, I was genuinely wrong more than half of the times…

I'll let @zner0L decide whether we want to crank up the quality.

mal-tee:
I agree, its not that important for most images. Maybe in the future I will just add an option for higher quality to some thumbnails.

baltpeter (approved):
Thank you!

PR #693, opened 2021-09-23
Draft: fix #545

### LINKED ISSUE #545: Suggest: Almost unusable on mobile
The input fields are too small and the whole layout isn't mobile friendly.

mal-tee:
This works, but messes up the item lists. Any tips on how to circumvent that?

image
image

### DIFF
--- a/assets/styles/main.scss
+++ b/assets/styles/main.scss
@@ -782,6 +782,9 @@ body.sticky-offset {
 }
 
 .brutusin-form {
+    @include mobile {
+        line-height: 1.3;
+    }
     table.object {
         width: 100%;
     }
@@ -808,13 +811,18 @@ body.sticky-offset {
         vertical-align: top;
 
         &.prop-name {
-            width: 300px;
+            @include desktop {
+                width: 300px;
+            }
         }
         td.prop-name {
             width: auto;
         }
         &.item-action {
             width: 35px;
+        }
+        @include mobile {
+            display: block;
         }
     }
 


assets/styles/main.scss
```diff
@@ -808,14 +811,19 @@ body.sticky-offset {
...
+            }
         }
         td.prop-name {
             width: auto;
         }
         &.item-action {
             width: 35px;
         }
+        @include mobile {
+            display: block;
```
baltpeter, line 825: You don't want to apply that to any `td` (`.prop-value` `td` contains another `table` with its own `td`s, which causes the messed up styling). Instead, you want to scope this to `td.prop-value` and `td.prop-name`:

image

assets/styles/main.scss
```diff
@@ -782,6 +782,9 @@ body.sticky-offset {
 }
```
baltpeter, line 782: Now that one field/row occupies multiple lines, you need to increase the spacing between the rows. Otherwise, it's not obvious which input each label is for.

baltpeter (changes requested):
Sorry for taking so long with the review. :| This is actually quite easy to fix, see below.

assets/styles/main.scss
mal-tee: How could exclude the nested tables from this? :D

mal-tee: Otherwise everything would be spaced greater, which wouldn't benefit anyone.

assets/styles/main.scss
mal-tee: Thanks!

assets/styles/main.scss
baltpeter: `tr:not(.item)` should work.

baltpeter: Maybe you need to extend the `:not()`. I haven't checked whether there are any other nested tables.

PR #700, opened 2021-09-25
Objection template: Ask for prior notification in case of deletion

baltpeter:
See datenanfragen/data#1335.

This does make the print versions of those templates quite ugly unfortunately, but what can you do…


mal-tee (approved):
Thank you! Consider this approved if the tests don't fail.

PR #706, opened 2021-10-01
List press and research mentions

baltpeter:


mal-tee:
I think we should add the research section to the German page as well. 😄

zner0L (changes requested):
Did you leave out the research mentions on the German page because the paper is in English? I mean, there will hardly be any research published in German, so I think it is fine to put it on the German site as well. Otherwise, this seems very good to me, thank you!

baltpeter:
> Did you leave out the research mentions on the German page because the paper is in English?

Pretty much, yes. I briefly considered extracting them into Hugo data files and then automatically generating a "mentions in website language" and a "mentions in other languages" (we do even have a couple in languages other than German and English) section. But that seemed a bit overkill. :D 

But I can add the English research to the German page, sure.

baltpeter:
Done. Now it's your turn to write a couple of press releases so that section doesn't look so embarrassingly empty, @zner0L. :D

PR #708, opened 2021-10-02
Optimising deploy script to use GNU parallel

### LINKED ISSUE #674: Optimize deploy script
If you know your way around shellscripts, feel free to optimize our deploy.sh-script:
https://github.com/datenanfragen/website/blob/master/deploy.sh

We assume that time savings could be achieved by using parallelization, e.g. with GNU parallel. At the moment we copy a lot of stuff in simple loops:
https://github.com/datenanfragen/website/blob/4a2693b65db13d182ca8ae4dc3f58a19ea0b313f/deploy.sh#L56-L60

mohitsaxenaknoldus:
Fixes #674


mohitsaxenaknoldus:
I need some help fixing this Netlify thing. It says it can't find `apk`.

mal-tee:
I think we have to install that manually. @baltpeter Can you do us the honors? 

But in general we want to move the installation outside of the `deploy.sh` script. Mohit, can you please add parallel as a requirement to the README and add it to our CI config: https://github.com/datenanfragen/website/blob/master/.circleci/config.yml

mal-tee:
Can you try it again without the install-step? It might be preinstalled in netlify.

mohitsaxenaknoldus:
> Can you try it again without the install-step? It might be preinstalled in netlify.

Looks like it doesn't.

mal-tee:
Sad. CI stuff. :( That means we have to check if we are in netlify and if we are, we should obtain a binary from somewhere and just use that. Or build it ourselves in the script.

mal-tee:
I've just tested the improved script and I am quite happy:

```
# old:
./deploy.sh  218,96s user 28,39s system 191% cpu 2:09,35 total
# this PR:
./deploy.sh  221,53s user 19,38s system 250% cpu 1:36,34 total
```

deploy.sh
```diff
@@ -53,11 +53,14 @@ cd content || exit
...
-do
+fun(){
+    lang=$1
     find "$lang/company" -name '*.json' -exec sh -c 'mv "$0" "${0%.json}.md"' {} \;
     find "$lang/supervisory-authority" -name '*.json' -exec sh -c 'mv "$0" "${0%.json}.md"' {} \;
-done
+}
+
+export -f fun
+parallel fun ::: de en fr pt es hr
```
baltpeter, line 63: Would it be possible to use the `languages` array we have already instead of hardcoding the values again?

deploy.sh
```diff
@@ -53,11 +53,14 @@ cd content || exit
 # Unfortunately, Hugo only accepts .md files as posts, so we have to rename our JSONs, see https://stackoverflow.com/a/27285610
 echo "Renaming JSON files…"
 
-for lang in ${languages[@]}
-do
+fun(){
```
baltpeter, line 56: Let's use a more descriptive name like `rename_for_hugo`.

baltpeter (changes requested):
Thank you! Just some comments from me.

baltpeter:
> I've just tested the improved script and I am quite happy:
> 
> ```
> # old:
> ./deploy.sh  218,96s user 28,39s system 191% cpu 2:09,35 total
> # this PR:
> ./deploy.sh  221,53s user 19,38s system 250% cpu 1:36,34 total
> ```

Not sure whether this will yield a noticeable improvement, but it would be nice if we could benchmark whether also parallelizing the copying of the companies and supervisory authorities is worth it.

deploy.sh
```diff
@@ -34,11 +38,14 @@ cp data_tmp/companies/* static/db
 cp data_tmp/suggested-companies/* static/db/suggested-companies
 cp data_tmp/supervisory-authorities/* static/db/sva
 
-for lang in ${languages[@]}
-do
+copy_file(){
```
baltpeter, line 41: Please rename that to `copy_records`.

deploy.sh
```diff
@@ -18,11 +18,15 @@ echo "Fetching data…"
...
+
+create_directory(){
+    lang=$1
     mkdir -p "content/$lang/company"
     mkdir -p "content/$lang/supervisory-authority"
-done
+}
+
+export -f create_directory
+parallel create_directory ::: ${languages[@]}
```
baltpeter, line 29: You don't need to do that for this. Seems a little overboard for creating a couple of directories. :D

baltpeter (changes requested):
Thanks! Other than the comments below, if you install `parallel` in Netlify, this is good to go as far as I'm concerned.

mohitsaxenaknoldus:
> Thanks! Other than the comments below, if you install `parallel` in Netlify, this is good to go as far as I'm concerned.

Resolved the comments. How can I install `parallel` in Netlify?

baltpeter:
That's unfortunately not so easy. You don't have root in the container (https://answers.netlify.com/t/apt-get-install-build-dependency/3095/2) and there doesn't seem to be a binary distribution of `parallel` (https://www.gnu.org/software/parallel/).

And compiling it each time would likely negate the benefits of using `parallel`. But you could try this:

https://git.savannah.gnu.org/cgit/parallel.git/tree/README#n84

mohitsaxenaknoldus:
> That's unfortunately not so easy. You don't have root in the container (https://answers.netlify.com/t/apt-get-install-build-dependency/3095/2) and there doesn't seem to be a binary distribution of `parallel` (https://www.gnu.org/software/parallel/).
> 
> And compiling it each time would likely negate the benefits of using `parallel`. But you could try this:
> 
> https://git.savannah.gnu.org/cgit/parallel.git/tree/README#n84

Is it not possible to install it on your Netlify runner from the backend?

baltpeter:
> Is it not possible to install it on your Netlify runner from the backend?

No.

mal-tee:
I was slightly mistaken. We need to install it manually as in the deploy-script, cause we can not install things. Did you try to acquire a binary, as @baltpeter proposed?

mohitsaxenaknoldus:
> I was slightly mistaken. We need to install it manually as in getting it in the deploy-script, cause we can not install things. Did you try to acquire a binary, as @baltpeter proposed?
> 
> > https://git.savannah.gnu.org/cgit/parallel.git/tree/README#n84

That part is done! But to install the binary we need a `dpkg` installed and for that we need to have `apk` on Netlify.

mohitsaxenaknoldus:
Ok so after adding `exit` to the loop it is now missing the `public` folder which is being ignored by the `.gitignore`.

mohitsaxenaknoldus:
> Ok so after adding `exit` to the loop it is now missing the `public` folder which is being ignored by the `.gitignore`. Is it supposed to fail at this point guys? I think we are close!

mohitsaxenaknoldus:
> > Which Ubuntu container is it using? @mal-tee @baltpeter
> 
> Seems like netlify are using 20.04 (https://github.com/netlify/build-image/blob/focal/Dockerfile)
> 
> > Ok so after adding exit to the loop it is now missing the public folder which is being ignored by the .gitignore. Is it supposed to fail at this point guys?
> 
> No.

It was missing the `public` folder due to this branch filter: `if [ "$NETLIFY" = "true" ] && [ "$BRANCH" = "master" ];`
I've fixed it now! You can re-run your performance tests again @mal-tee @baltpeter

mohitsaxenaknoldus:
> As @mal-tee said, disabling the deploy previews with Netlify is obviously not an option.
> 
> I will look at solving the Netlify problem myself on your branch. But I can't make any promises as to when that is going to happen, there's a lot of other stuff I have to do first.

Hey, were you guys able to fix it?

baltpeter:
> Hey, were you guys able to fix it?

No, unfortunately, I didn't have time for that yet.

mal-tee:
Seems like parallel is (or was?) a single perl script: https://www.gnu.org/software/parallel/parallel_design.html

PR #709, opened 2021-10-02
Update AUTHORS

mohitsaxenaknoldus:


baltpeter:
Please just do that in your existing PR. No need to open an additional one.

PR #712, opened 2021-10-02
README: Mention necessary Hugo version

baltpeter:
We've recently seen a few problems where people were getting build failures due to using a Hugo version that's too old.


mal-tee (approved):
Thank you!

Issue #640, opened 2021-08-15
Donate form doesn't accept comma as decimal seperator


mal-tee:
It seems like the donate form does not accept a comma as decimal separator. This is quite unfortunate, as it's the norm in Germany.

![comma_sep](https://user-images.githubusercontent.com/45886910/129491695-593f03b2-4d28-438f-a476-b37fba148a12.gif)

In this example I was using the numpad on a german keyboard.

kaggrwal:
@mal-tee I would like to work on this, can you please guide me through the codebase?

mal-tee:
https://github.com/datenanfragen/website/blob/master/src/Components/DonationWidget.js

This is the code for the `/donate` page. The bug should be hidden here somewhere. Part of fixing a bug is understanding why it's happening. Only then you can fix it.

If you want to know how to get our project running locally, please read the README carefully.

kaggrwal:
Can you assign this to me?

kaggrwal:
I have raised a PR, please review it #714

PR #714, opened 2021-10-02
fixes for comma as decimal separator

kaggrwal:
Fixes for #640

### DIFF
--- a/src/Components/DonationWidget.js
+++ b/src/Components/DonationWidget.js
@@ -84,13 +84,11 @@ export default class DonationWidget extends Component {
                         <div className="input-addon-container">
                             <input
                                 id="donation-widget-amount"
-                                type="number"
-                                step="0.01"
-                                min="0.00"
+                                type="text"
                                 className="form-element"
                                 style="text-align: right;"
                                 value={this.state.amount}
-                                onInput={(e) => this.setState({ amount: Number.parseFloat(e.target.value) || 0 })}
+                                onInput={(e) => this.setState({ amount: this.preformatFloat(e.target.value) || 0 })}
                             />
                             <div className="input-addon">€</div>
                         </div>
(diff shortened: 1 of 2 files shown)


src/Components/DonationWidget.js
```diff
@@ -84,13 +84,11 @@ export default class DonationWidget extends Component {
                         <div className="input-addon-container">
                             <input
                                 id="donation-widget-amount"
-                                type="number"
```
mal-tee, line 87: We don't really want to change that.

kaggrwal: But you can't identify the comma then, It can only be taken as input when it is text.

baltpeter: I've dug into this a little: The `number` input isn't a problem. If you set the locale to German in Firefox, a regular `<input type="number">` accepts commas just fine. Somehow our code breaks this.

mal-tee: seems like you are right: https://jsfiddle.net/r9L7bj4a/7/

baltpeter: You can actually even debug this without setting your browser to German: Even in English, entering a decimal with a period doesn't work. It's likely that that's the same problem.

kaggrwal: I tried changing the locale in google chrome, but I couldn't input a comma.

baltpeter: Works fine for me in both Firefox and Chrome:

https://user-images.githubusercontent.com/4048809/135908901-348b3d73-03fa-4d65-97d5-62f87128d365.mp4

kaggrwal: Are you using any extension ?
 I used some extension to change the locale on my google chrome.

baltpeter: No, that is a native installation on a German system. The extension probably isn't the right way to change the locale.

But as I said, for the issue, you can focus on the period as the decimal separator.

davenewham: The reason the website breaks decimal input is that it is parsed as a float on input, so "10," gets cut to 10 meaning that you can never add a decimal, but if you use the stepper, the input becomes "10,01" which is valid and is then parsed correctly.

Other than removing this parsing ball together, I think it's the correct way of doing it. The || 0 needs to be omitted. However, this code chunk comes from StackOverflow so will need correct attribution

kaggrwal: So, you mean you just want to treat period "."  as a decimal operator. 
Correct me if am wrong, but the period is already treated as a decimal separator.

baltpeter: No, it's not. If you try to enter a decimal point in the donate form, it is silently discarded.

baltpeter:
Hey @kaggrwal, are you still working on this or do you want me to close the PR and unassign you from the issue?

Issue #532, opened 2021-01-18
SVA-Finder: Specify special cases for the Bundesbeauftragten


mal-tee:
e.g. telecommunication companies

baltpeter:
The BfDI is responsible in particular for ([1](https://stiftungdatenschutz.org/fileadmin/Redaktion/Dossiers_Infoplattform/2020-07-Update/DSGVO-Praxis__Aufsichtsbehoerden_Zustaendigkeiten_Juni2020.pdf)):

* öffentliche Stellen des Bundes
* Telekommunikations- und Postdienstunternehmen
* private Unternehmen, die unter das Sicherheitsüberprüfungsgesetz fallen
* Jobcenter

baltpeter:
I'll put this on hold until datenanfragen/data#1126 is done as we'll also need to account for the Rundfunkdatenschutzbeauftragten in the SvaFinder.

baltpeter:
See also: https://www.bfdi.bund.de/DE/Service/Kontakt/Kontaktfinder/kontaktfinder_node.html

moriarty-jim:
I'm preparing to work on this when datenanfragen/data#1126 is merged. If I understand your code structure correctly, changes need to be made to:

`src/Components/SvaFinder.js` something like this:

```javascript
const STEPS = {
  country: {
    de: {
      // ...
      rundfunk: {
        'rundfunk-dw': 'derfdw',
        // ... (add all broadcasters)
      },
      // ...
    },
    // ...
  },
};
// ...
const SVAS = {
  derfdw: 'Rundfunkdatenschutzbeauftragter Deutsche Welle',
  // ... (add all Rundfunkdatenschutzbeauftragte)
}
```

and all the translation files in `src/i18n/` for languages in use (de, en, fr, hr, pt, es):

```
{
...
    "sva-finder": {
        "rundfunk": "Öffentlich-rechtlicher Rundfunk",
        "rundfunk-q": "Auf welche öffentlich-rechtliche Rundfunkanstalt bezieht sich Dein Anliegen?",
        "rundfunk-dw": "Deutsche Welle",
        ...
     }
...
}
```
Anywhere else? Or am I completely on the wrong track?

baltpeter:
> I'm preparing to work on this when datenanfragen/data#1371 is merged. 

Awesome! I'll assign you in that case.

> Anywhere else? Or am I completely on the wrong track?

No, that should be it. Feel free to ask if anything's unclear but looks like you've already figured it out.

[...]

PR #717, opened 2021-10-03
SVA-Finder: Special cases for Rundfunkdatenschutzbeauftragte (public broadcasting SVAs)

moriarty-jim:
resolves partly #532


baltpeter:
Thanks, @moriarty-jim! This looks good and is essentially good to be merged.

It would be great if you could add a quick test for this, though. This essentially just means adding another path (with a list of the steps (they all start from Germany) and the expected result) here:

https://github.com/datenanfragen/website/blob/f0c13e3340ea4a2f038f53f9e2cb7607a68978af/cypress/integration/components/svafinder.spec.js#L23-L69

Just one path for one of the broadcasters is enough as they are all pretty much the same.

moriarty-jim:
Done!

baltpeter (approved):
Great, thank you!

PR #719, opened 2021-10-03
Draft: server sider render comments

mal-tee:
This is still a bit rough, but intended as a proposal: Do we want sth. like this? Or does this change to much stuff with too little benefits?


baltpeter:
Thanks, this is an interesting idea! I would like to have something like this. A few unordered notes/questions:

* At the moment, for JS users, it immediately refetches the comments on page load. We had some previous discussion on optimizing this. Do you plan on doing that?
* I don't think it requires too many change (at least in the current state).
* The build speed increase should be manageable. At the moment, you have a pretty big and unnecessary overhead from using `ts-node` which first has to do type checking every time.
* As far as I understand it, this isn't "really" SSR, which would include "rehydration" later on. But that brings its own problems and I think I actually much prefer your approach in this particular case.

Issue #679, opened 2021-09-20
Record failed production tests in Cypress


baltpeter:
We have a test configured in Circle CI that automatically tests our production site on a schedule:

https://github.com/datenanfragen/website/blob/f9cc1f0c88c1f1768e3b418efcabb0aed80cf807/.circleci/config.yml#L97-L111

Unlike our "regular" tests, this doesn't record the runs to the Cypress dashboard. We should record failed production tests as well to ease debugging them. 

This should be as simple as adding `--record` to `yarn cypress run` if the `CYPRESS_RECORD_KEY` env var is set and the run failed.

imrishabh18:
I would like to take this up

mal-tee:
Happy hacking, I will assign you.

AnirbanBall:
I would like to work on this

mal-tee:
Sorry, @imrishabh18 Is already working on this.

PR #720, opened 2021-10-03
chore: records the production test in dashboard

imrishabh18:
Closes - #679

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,7 +108,7 @@ jobs:
             command: |
               yarn
               yarn cypress verify
-              yarn cypress run
+              yarn cypress run --record
 
 workflows:
   version: 2


baltpeter (changes requested):
Thanks for the PR, @imrishabh18. But please read the issue carefully. We only want to record the run under certain conditions, not always.

imrishabh18:
@baltpeter I have added the condition to check if the env is set or not. What should be the condition to check if the build passed or not?

baltpeter:
Now that I've looked into this further, not sure, actually.

There is a `videoUploadOnPasses` option (https://github.com/cypress-io/cypress/issues/2217#issuecomment-410386771) but I am not sure whether that still records the run without videos. That would not be an option as we have a limited number of recordings available in the Cypress dashboard and the automated tests would soon take that up completely. What do you think, @zner0L?

zner0L:
Reading [the documentation](https://docs.cypress.io/guides/guides/screenshots-and-videos#Videos), this seems like the correct option to set and would suggest trying. If that fails, I am also fine with getting the videos as artifacts out of CircleCI.

PR #722, opened 2021-10-04
adds simple tooltip to the comment label

duaimei:


src/Components/CommentsWidget.js
```diff
@@ -259,7 +259,17 @@ export class CommentForm extends Component {
                 <div className="clearfix" />
 
                 <div className="col25 col100-mobile">
-                    <strong>
+                    <strong
+                        title={`The following tags are supported:
```
duaimei, line 263: Per the request in the issue. This is the most subtle way of adding a tooltip to this widget. 

If the comment needs to be internationalized, then I request help in how to go about that.

I did not see any tooltip libraries being used in this repo, but but we can use one if you wish, or add a (i) icon. (but an icon isn't very subtle)

duaimei:
Screen Shot 2021-10-03 at 9 26 06 PM

This is what the tooltip looks like

src/Components/CommentsWidget.js
mal-tee: Thank you! I think this kind of tooltip is sufficient, but I think it would make more sense if the tooltip would appear on the field it self as well. 

Use the `t()` function for internationalisation, like here: 

https://github.com/duaimei/datarequests-website/blob/8c54b4145f14486b6b9df238360169b55046f391/src/Components/CommentsWidget.js#L53

 and add the string in `src/i18n/en.json`

duaimei:
I just wanted to let you know that I have seen your comments, and I will work on it this weekend!

baltpeter:
Hey @duaimei, are you still working on this or do you want me to close the PR and unassign you from the issue?

Issue #641, opened 2021-08-15
Testing: Error handler


mal-tee:
We have a handy error handler, showing a modal to users asking them to report an error to us. 

![image](https://user-images.githubusercontent.com/45886910/129492268-17aebf66-6c20-4763-9a69-fd2029d768d0.png)

At the moment we only have very basic testing for it: https://github.com/datenanfragen/website/blob/master/cypress/integration/error-handler/console-logging.spec.js

We would like to have some more tests in cypress covering the error handler:

- [ ] Is the modal shown correctly?
- [ ] Do the `mailto `and GitHub links work correctly?
- [ ] Does the "De-JSON-ification" work?
	- we have to make the error dump look as little as JSON as possible, to circumvent email spam filters 😅 
	- see the collapsed example below

<details>
<summary>📖 Example for De-JSON-ification</summary>

```js
{
    "message": "Error: Browsers are awesome!",
    "type": "error",
    "colno": 19,
    "filename": "javascript:(function(){throw new Error(\"Browsers are awesome!\")})();",
    "lineno": 1,
    "error": {
        "message": "Browsers are awesome!",
        "name": "Error",
        "stack": "@javascript:(function(){throw new Error(\"Browsers are awesome!\")})();:1:19\n@javascript:(function(){throw new Error(\"Browsers are awesome!\")})();:1:55\n"
    },
    "defaultPrevented": false,
    "eventPhase": 2,
    "isTrusted": true,
    "returnValue": true,
    "code_version": "1.0.0",
    "user_agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:91.0) Gecko/20100101 Firefox/91.0",
    "url": {
        "href": "https://www.datarequests.org/",
        "origin": "https://www.datarequests.org",
        "protocol": "https:",
        "host": "www.datarequests.org",
        "hostname": "www.datarequests.org",
        "port": "",
        "pathname": "/",
        "search": "",
        "hash": ""
    }
}
```

becomes 

```
**Debug information:**

message := Error: Browsers are awesome!
type := error
colno := 19
filename := javascript:(function(){throw new Error("Browsers are awesome!")})();
lineno := 1
error.message := Browsers are awesome!
error.name := Error
error.stack := @javascript:(function(){throw new Error("Browsers are awesome!")})();:1:19
@javascript:(function(){throw new Error("Browsers are awesome!")})();:1:55

error.context := defaultPrevented := false
eventPhase := 2
isTrusted := true
[...]

PR #724, opened 2021-10-04
Testing: Error Handler 

InfiniteVerma:
Fixes: #641 

- [x] Is the modal shown correctly?
- [x] Do the mailto: and GitHub links work correctly?
- [ ] Does the "De-JSON-ification" work? 
 
Following the new comments on the issue, I was able to throw the custom error with error code specified. But it never happens if I remove the line below: 

```javascript
doc.location = 'javascript:(function(){throw new Error()})();';
```

Subsequently, the error from the above function is shown on the logs and the same is sent for De-JSON-ification.

### DIFF
--- a/src/test-interface.js
+++ b/src/test-interface.js
@@ -2,6 +2,7 @@
 
 import FlashMessage, { flash } from 'Components/FlashMessage';
 import localforage from 'localforage';
+import { CriticalException } from './Utility/errors';
 
 window.showFlash = function (type, text, duration) {
     flash(
@@ -20,4 +21,8 @@ window.accessLocalForageStore = function (store_name) {
             })
         );
     });
+};
+
+window.throwError = function () {
+    return new CriticalException();
 };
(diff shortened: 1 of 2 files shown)


src/test-interface.js
```diff
@@ -21,3 +22,7 @@ window.accessLocalForageStore = function (store_name) {
         );
     });
 };
+
+window.throwError = function () {
+    return new CriticalException();
```
baltpeter, line 27:
```suggestion
    throw new CriticalException();
```

Also, it would be good if you passed the parameters (`message`, `context`, `enduser_message` and the actual error obj) and then checked whether they get handled correctly.

baltpeter:
> Following the new comments on the issue, I was able to throw the custom error with error code specified. But it never happens if I remove the line below

Interesting. In Firefox, you can get around this problem by wrapping the throwing in a `setTimeout` with a duration of `0`. In Chrome, that doesn't seem to work.

Anyway, in that case you can continue to use the `window.location` hack with your `throwError()` function (which will work with the fix I suggested above).

Also, please don't override the existing console logging test but create a new one.

baltpeter:
Hey @InfiniteVerma, are you still working on this or do you want me to close the PR and unassign you from the issue?

InfiniteVerma:
@baltpeter Sorry for the inactivity...yes I want to finish up the remaining things in the issue.

PR #725, opened 2021-10-04
Fix "GDPR", "RGPD", and "DSGVO" misspellings

baltpeter:
For an organization focussed almost entirely on this one legislation,
we sure misspell it a lot. ;)

I went through all possible permutations, so this should hopefully
catch them all (for now, anyway…).

image


mal-tee (approved):
Great 😆 Thanks.

PR #731, opened 2021-10-05
Add Outlook Webmail Mailto-Link

moriarty-jim:
see also: #402 
source for URL scheme: [https://stackoverflow.com/a/69080263](https://stackoverflow.com/a/69080263)

I tested it with a few generated letters and it seems to work fine. Additional and more structured testing would probably be a good idea.
Translations for "outlook" are still there.


zner0L (changes requested):
Hm, so when I try to use your link I get a runtime error:
Screenshot 2021-10-06 at 17-01-08 Runtime Error

What are you doing differently so that it works for you?

zner0L:
I guess you didn' hear back from them. As long as this is broken, I don't see a reason to keep this open anymore. If you have new change, feel free to reopen.

PR #732, opened 2021-10-06
Suggest  page: Alphabetically sort relevant countries

### LINKED ISSUE #551: Suggest: Sort relevant countries list alphabetically

Cool-Runningz:
This PR contains the fix to alphabetically sort the relevant countries list on the "Suggest" page.

Fixes #551

### DIFF
--- a/src/suggest-edit.js
+++ b/src/suggest-edit.js
@@ -30,7 +30,16 @@ window.onload = () => {
         });
 };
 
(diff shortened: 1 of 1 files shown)


baltpeter (changes requested):
Thanks, @Cool-Runningz. This looks good.

As @mal-tee suggested in #551, I think it would be good to do the same thing for the categories as well.

PR #733, opened 2021-10-06
Fix the race conditions in the production test

zner0L:
This is the third try of fixing the race conditions in the production tests. I attempt that by doing two things: Treating the state asynchronously by properly using callbacks like we should and actually passing the current state to the modal, so they actually reflect the state they are supposed to.

Please review carefully because I might screw up something that depended on or timing being so messy.

### DIFF
--- a/src/Components/RequestGeneratorBuilder.js
+++ b/src/Components/RequestGeneratorBuilder.js
@@ -91,47 +91,54 @@ export default class RequestGeneratorBuilder extends Component {
                 return prev;
             });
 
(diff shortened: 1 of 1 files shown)


baltpeter (changes requested):
Thanks, this makes me sad. :( :D 

Changes look sensible to me and the tests pass. I'll do a bit of manual testing but I think this should be fine.

baltpeter:
Couldn't find any more problems.

baltpeter (approved):
Thank you!

PR #740, opened 2021-10-07
fixed external links UI issue

### LINKED ISSUE #668: Broken external link styling on open source page
On the [open source page](https://www.datarequests.org/open-source), the external link styling is broken in various ways:

image

Not sure what to do here.

The most obvious problems are that the indicator is in the next line for the icons and headings. That's because those are block elements which push their `::after` (which is used to insert the indicator) to the next line. You _could_ "solve" the problem by adding the `no-link-decoration` class to the links but now users wouldn't have any indication that those are external links, which means that isn't an option.  
Another option (which I'm not a fan of either but it would be less unacceptable) would be to just remove the links from the images and headings…

Either way, the "View repository" buttons definitely also need external link styling.

tiger-yash:
**Fixes**

Fixes #668 by @baltpeter 

**Description**

Made the desired UI changes to fix external link tag issue

### DIFF
--- a/assets/styles/main.scss
+++ b/assets/styles/main.scss
@@ -595,20 +595,6 @@ table#my-requests-table {
             hyphens: auto;
         }
     }
-    &>.button-link{
-        text-decoration:none;
-        background-position: bottom left;
-        background-repeat: no-repeat;
-        background-size: 100% 2px;
-        background-size: calc(100% - 16.8px) 2px !important;
-        color:white !important;
-        &:after {
-            color: white !important;
-            font-icon: 80% url('external.svg') !important;
-            padding-left: 4px !important;
-            vertical-align: -0.5px !important;
-        }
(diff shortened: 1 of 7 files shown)


tiger-yash:
@zner0L 
I have made the appropriate changes, please review

tiger-yash:
@baltpeter @zner0L 
Please review

zner0L:
Not only does this PR repeatedly miss the mark of what we requested in the issue and the discussion, the author also told us via e-mail they would retract the PR if we wouldn't review it by Sunday. In consequence, I will close this PR.

@tiger-yash, please keep in mind that lots of open source projects, this one included, are maintained by volunteers, who are especially busy during Hacktoberfest and can therefore sometime not react within hours or days to your changes. I hope you are more patient with other open source projects.

PR #745, opened 2021-10-09
Suggest Company: Validate if company already exists

### LINKED ISSUE #546: Suggest: "We already have these similarly named companies: <>"
Not really the right place for this but with regards to datenanfragen/data#970, it would be really nice if `website` had a "We already have these similarly named companies: <>" kind of thing.

_Originally posted by @baltpeter in https://github.com/datenanfragen/backend/issues/30#issuecomment-770997423_

daniel-shuy:
Fixes #546 and datenanfragen/backend#30

A company is considered duplicate if the generated slug is the same (the slug is generated from the website or name).

I'll need someone to verify the translations, I generated them using Google Translate.

image

### DIFF
--- a/cypress/integration/use-cases/suggest.spec.js
+++ b/cypress/integration/use-cases/suggest.spec.js
@@ -8,21 +8,25 @@ describe('Using the suggest form', () => {
         // so lets do it manually
         cy.window().then((win) => {
             win.fetch = async function (x, opt) {
-                if (!win.fetchlog) {
-                    win.fetchlog = [[x, opt]];
+                if (opt && opt.method === 'PUT') {
+                    if (!win.fetchlog) {
+                        win.fetchlog = [[x, opt]];
+                    } else {
+                        win.fetchlog.push([x, opt]);
+                    }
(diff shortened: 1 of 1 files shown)


baltpeter:
Thanks for the PR, @daniel-shuy! Sorry that the issue doesn't go into more details (it was originally meant for Hacktoberfest) but here's what I had in mind:

* To check for similarly named companies, we should use our search engine (Typesense) which can is already typo-tolerant and can do fuzzy matching. You can see how to access the search results from that in [`SearchBar.js`](https://github.com/datenanfragen/website/blob/master/src/Components/SearchBar.js).
* Instead of hard-failing in the case of matches, I think we should just display them to the user under the name field and link to their company pages so they can check whether this is a different company or not. Either way, they should still be allowed to submit.

The inspiration for this feature are contact forms that suggest FAQ items based on the subject you enter, like this one for Humble Bundle:

image

Also, you can just remove the machine-translated strings. As long as you fill in English, our translators will add the missing translations later through Weblate.

Issue #743, opened 2021-10-08
Use Cypress 7/8 capabilities


mal-tee:
- [ ] test downloads
- [ ] use intercept in suggest-test
	- https://github.com/datenanfragen/website/blob/652baf94bfec398a72ca5acb39016392d110b2a5/cypress/integration/use-cases/suggest.spec.js#L7-L10

PR #748, opened 2021-10-09
Cypress 8 update tests

davenewham:
Use Intercept to mock the request

Use the new download feature to download the PDF from the website, and at least assert it is not empty

As advised in #743


davenewham:
Closes #743

davenewham:
Baseline test results: 65 passed, 2 failed (v8.5.0) 
Request language fallback (request-language-fallback.spec.js)
Changing the country should update the companies in the wizard (country-change-updates-wizard.spec.js)

PR #782, opened 2021-10-14
Make Hugo footnotes more accessible

### LINKED ISSUE #239: Make Hugo footnotes more accessible
See: https://danielpost.com/articles/making-hugos-footnotes-accessible/

Cool-Runningz:
This PR contains the code to modify the Hugo footnotes to make them more accessible and user friendly.  Fixes #239 

## Implementation Details 
- I adapted the CSS & HTML from the [Hacker News comment](https://news.ycombinator.com/item?id=24747376) shared within the [issue](https://github.com/datenanfragen/website/issues/239#issuecomment-932588187). 
- I currently have things setup to run the `footnotes.js` script when viewing a page within the `/blog` directory. I looked through a majority of the pages throughout the website and only found footnotes on this one [blog post](https://www.datarequests.org/blog/honey-data-collection). Based on that I assumed that blog posts would be the main candidate for checking for footnotes but I can change it if that is not the desired behavior.
-  I noticed that Hugo renders the footnotes embedded within the page with an ID pattern of `fnref:<index>` where the index corresponds to the sequential order of the footnote. The bottom footnotes are rendered with an ID pattern of `fn:<index>`. Based on that I grabbed all the embedded footnotes with the matching ID pattern, took the necessary info I needed from it, and then replaced and removed the old footnotes. 
   - I tried to make the changes for this feature as unobtrusive as possible. Having things structured this way made it so that I didn't have to add anything to the MD files. My hope is that this makes things easier.
   - I made sure to keep the anchor links intact so that if you are viewing a footnote at the bottom and click on the ⬅️   it will scroll you back to the location within the page that it refers to.


Cool-Runningz:
@baltpeter  I believe I have addressed all the comments so this should be ready for another review.  Thank you very much for all your feedback 🙂 

As suggested by you and @zner0L in [this comment](https://github.com/datenanfragen/website/issues/239#issuecomment-940511904), I went ahead and relocated the footnote rendering logic into `general.js` and deleted the `footnote.js` file. I did notice that the CircleCi tests started failing after doing that and I'm not sure why that is happening since everything else passes 🤷🏽‍♀️

baltpeter:
Thank you! Other than the test failures, this is now good to merge as far as I'm concerned.

(Still waiting on @zner0L's a11y review, though…)

baltpeter (dismissed):
Thank you!

Cool-Runningz:
Thanks for the feedback! I will work on addressing the requested changes and see what I can do in regards to setting the appropriate aria attributes. I will try and make those updates sometime within the week.

Cool-Runningz:
@zner0L I have addressed the change requests by removing the `text-underline` style and adding logic to close the details pop-up when a user clicks outside of the element. I came up with a slightly different solution than what was suggested to accommodate the accessibility improvements.

Within the `<summary>` element I added `<span className="visually-hidden">Footnote</span>` which will [visually hide the label while still making it accessible to screen readers](https://www.w3.org/WAI/tutorials/forms/labels/#note-on-hiding-elements). I tested this out in VoiceOver and instead of just reading out the number it will now pronounce something like _"**Footnote 4**, collapsed disclosure triangle"_. In addition, when viewing the [VoiceOver rotor](https://support.apple.com/guide/voiceover/with-the-voiceover-rotor-mchlp2719/mac), the same text content is displayed. In my opinion, this is a sufficient indicator to the screen reader user that the `details` element they have landed on is actually a footnote with content. 

The only con is that it is hardcoded in English but I wasn't sure of where to add a translation key so assistance there would be appreciated.  Anyway, I just wanted to propose that accessibility solution and you can let me know what you think. Thanks!

assets/styles/elements.scss
```diff
@@ -765,3 +765,17 @@ details.footnote {
         }
     }
 }
+
+/* Adapted after: https://www.w3.org/WAI/tutorials/forms/labels/#note-on-hiding-elements
```
baltpeter, line 769: You can use our existing class for that: https://github.com/datenanfragen/website/blob/93ecc4aa7ed846db3459e2e3f7be40e027d13ba9/assets/styles/framework.scss#L380-L389

src/Components/Footnote.js
```diff
@@ -23,6 +23,7 @@ const Footnote = (props) => {
     return (
         <details className="footnote" id={props.id} ref={detailsRef}>
             <summary>
+                <span className="visually-hidden">Footnote</span>
```
baltpeter, line 26: For translating this, you can add just the English translation (the other ones will later be filled in by our translators on Weblate) to this file: https://github.com/datenanfragen/website/blob/master/src/i18n/en.json  
The top-level keys set the `scope` (I don't think any of the existing ones fit, so you have to invent a new one), the second-level keys the translation ID.

Then, you wrap the component in an `IntlProvider` and specify the translation's `scope` :

https://github.com/datenanfragen/website/blob/802a2805a57f54a7b344b74f15eba347298e3f7a/src/Components/Generator/TransportMediumChooser.js#L10

Finally, you can use `<Text>` with your translation ID:

https://github.com/datenanfragen/website/blob/802a2805a57f54a7b344b74f15eba347298e3f7a/src/Components/Generator/TransportMediumChooser.js#L12

Cool-Runningz: @baltpeter Thank you for the assistance! That helped a lot and this should be ready for another review.

baltpeter (commented):
Thanks again, @Cool-Runningz! I'm really happy with how this turned out.

And since @zner0L doesn't seem to have any comments anymore either, let's get this merged.

baltpeter (approved):
Wrong button… :D

Cool-Runningz:
Thanks @baltpeter! It’s been fun getting to contribute during Hacktoberfest! Thank you all for the opportunity and hopefully I can contribute more in the future.

PR #789, opened 2021-10-17
Hide fixed fields with same value/desc

### LINKED ISSUE #735: Hide fixed fields with same value as the generator in the dropdown
On our [“My saved data” page](https://www.datarequests.org/id-data-controls/), you can enter your details for them to be prefilled in the generator. You can also add your own fields which is useful for example if you have multiple email addresses.

In the generator, we have a dropdown where you can manually add them to the identification data (we call them "fixed fields"):

https://user-images.githubusercontent.com/4048809/136290810-21ce7184-86d6-4a61-a7d1-7e59289e5960.mp4

But fields that are already present in the ID data are still displayed even though clicking them again won't do anything (or add them a second time, which doesn't make sense either):

https://user-images.githubusercontent.com/4048809/136290977-81f97abb-dd76-4c09-950a-c944783ff872.mp4

Instead, we should hide fields that are already present in the ID data (and have the same value!).

Jaagrav:
https://user-images.githubusercontent.com/52719271/137610475-b2da2570-65b4-4d94-bbdd-97e7cac12b34.mp4

Closes #735 

Currently it only checks if the field descriptions and types are equal or not. In case of addresses, there's an object that I could check if they are equal or not but then it doesn't make sense because I don't think people can store addresses with the same description on the [My Saved Data](https://www.datarequests.org/id-data-controls/) page.

@baltpeter @zner0L

### DIFF
--- a/src/Components/Generator/DynamicInputContainer.js
+++ b/src/Components/Generator/DynamicInputContainer.js
@@ -44,33 +44,45 @@ export default class DynamicInputContainer extends Component {
         // As this is at least the second time I have struggled to remember this: This is the button next to the 'add
         // new field' menu which allows you to add fields you have defined in the 'My saved data' section.
         let fill_fields = [];
+
         if (this.props.fillFields)
             this.props.fillFields.forEach((field) => {
-                fill_fields.push(
-                    <div className="fill-field">
-                        <div style="display: table-cell">
-                            {field.desc}:{' '}
-                            <span className="fill-field-value">
-                                {field.type === 'address'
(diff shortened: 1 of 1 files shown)


baltpeter (dismissed):
Thank you, @Jaagrav! I have left a few minor stylistic comments below.

As for whether the check is enough, we should ask @zner0L who is more familiar with this.

Jaagrav:
Please check now, if the requested changes you requested are added.

zner0L (approved):
Thank you!

PR #806, opened 2021-10-25
Dutch site

baltpeter:


mal-tee (commented):
Thank you!

We should update https://github.com/datenanfragen/website/blob/01f7ebab0d6701813931da039c4f0c8e456c5399/content/hr/verein/_index.md as well. After that we are good to go.

mal-tee (approved):

Issue #660, opened 2021-09-09
Generator: Long URLs in record comments overflow in mobile Chrome, causing off-screen modal positions


WebworkrNet:
**Steps to reproduce**
Using the wizard.
The last record to be called up was that of the "_Beitragsservice_".
The next data set is now to be selected.
**Expected behaviour**
As with other records, the attached note (in mobile view) should be in the user's view area.
**Actual behaviour**
Instead, for this record, the hint is displayed outside the field of view. Looks to me like in this case the design is not adapting to the device.
![Screenshot_20210909-223237_Firefox](https://user-images.githubusercontent.com/43509007/132762529-aae9894d-a16c-4b28-812d-516003c60960.jpg)

I have not tested Firefox.
![Screenshot_20210908-160152_Chrome](https://user-images.githubusercontent.com/43509007/132762602-ea5b03b5-eb42-479a-a400-628b26b2caea.jpg)

baltpeter:
I have confirmed this issue. It only occurs in mobile Chrome, not Firefox.

It can be reproduced by selecting the Beitragsservice in the generator (wizard is not necessary) and then triggering the "You haven't used your request." modal by clicking the "New request" button.

Behaviour in Chrome:

https://user-images.githubusercontent.com/4048809/133106865-edc0b4cc-bda3-4f0d-9d2f-0e59ccb791da.mp4

Behaviour in Firefox:

https://user-images.githubusercontent.com/4048809/133106936-1fec4a35-f0e2-4c23-ae4a-a44f5a458bb0.mp4

The problem is due to the long URL in the comments of the Beitragsservice record. In Firefox, this is wrapped correctly, but in Chrome it overflows. The modal _is_ correctly centered in the viewport but the viewport is too large due to the overflow. This is most likely the exact same issue I've fixed somewhere else in #636.

WebworkrNet:
I have understood the problem. However, I can't see any (big) difference in the two videos. Both look OK to me.

For me, it still looks different.

![Screenshot_20210914-223951_Chrome](https://user-images.githubusercontent.com/43509007/133331302-02898e4b-a873-42c9-9495-1f68f590d1ed.jpg)
![Screenshot_20210914-224009_Chrome](https://user-images.githubusercontent.com/43509007/133331308-a55b25a3-7081-448d-8a30-e1f9a5fde9dc.jpg)

**Device**: Samsung Galaxy XCover 4s (SM-G398FN)
**Operating System**: Android 11
[...]

PR #815, opened 2021-10-31
Issue 660 : Generator: Long URLs in record comments overflow in mobile Chrome

Rutam21:
This PR fixes https://github.com/datenanfragen/website/issues/660

### DIFF
--- a/assets/styles/generator.scss
+++ b/assets/styles/generator.scss
@@ -200,3 +200,7 @@ h2 {
 .request-transport-medium-chooser {
     margin-top: 17px;
 }
+
+#current-company-comments-colon {
+     overflow-wrap: anywhere;
+}


Rutam21:
@baltpeter Please review. Thanks

assets/styles/generator.scss
```diff
@@ -200,3 +200,7 @@ h2 {
 .request-transport-medium-chooser {
     margin-top: 17px;
 }
+
+#current-company-comments-colon {
```
baltpeter, line 204: This doesn't fix the issue as this not an actual ID that is in the output HTML. I would suggest looking at the page in your browser's dev tools to get a better sense of the HTML that is generated.

baltpeter (changes requested):

PR #822, opened 2021-11-10
Fixes #660: Enable wrapping for CompanyWidget comments in Chrome

### LINKED ISSUE #660: Generator: Long URLs in record comments overflow in mobile Chrome, causing off-screen modal positions

baltpeter:


mal-tee (approved):
Thank you!

PR #830, opened 2021-12-04
Better state management to (hopefully) fix the prod test

zner0L:
This is another try at fixing the productions tests. Since our concurrency model is a clusterf*ck that needs to be refactored to uses hooks and contexts to work properly, I can only try to patch up what I feel is responsible. I also cannot reproduce the problem locally so who knows if it works.

### DIFF
--- a/cypress/integration/old-bugs/request-saved-after-copymanually.spec.js
+++ b/cypress/integration/old-bugs/request-saved-after-copymanually.spec.js
@@ -1,0 +1,52 @@
+import { isOn, skipOn } from '@cypress/skip-test';
+/*
+ * In #830, we almost changed the behaviour that requests should be saved after the modal opened without waiting until the modal closed.
+ */
+
+describe('Request should exists without clicking ok in the copy manually modal', () => {
+    it('After click on the ActionButton', () => {
+        skipOn(isOn('production'));
+
+        cy.visit('/generator');
+        cy.contains('Information block').click();
+
+        cy.contains('Send email').click();
+        cy.get('.dropdown-container').contains('Copy text manually').should('be.visible');
+        cy.contains('Copy text manually').click({ force: true });
+
+        // Explicitly leaving out the click on the "Ok" button
+
+        cy.get('#reference')
+            .invoke('val')
+            .then((reference) => {
+                cy.window().then((win) =>
+                    win.accessLocalForageStore('my-requests').then((instance) => {
+                        instance.getItem(`${reference}-access`);
+                    })
+                );
+            });
+    });
+
+    it('After click in the new request modal', () => {
+        skipOn(isOn('production'));
+
+        cy.visit('/generator');
+        cy.contains('Information block').click();
+
+        cy.contains('New request').click();
+        cy.get('.modal').contains('Send email first').should('be.visible').click();
+        cy.get('.modal').contains('Copy text manually').click({ force: true });
+
+        // Explicitly leaving out the click on the "Ok" button
+
+        cy.get('#reference')
+            .invoke('val')
+            .then((reference) => {
+                cy.window().then((win) =>
+                    win.accessLocalForageStore('my-requests').then((instance) => {
+                        instance.getItem(`${reference}-access`);
+                    })
+                );
+            });
+    });
+});
--- a/cypress/integration/old-bugs/request-saved-after-copymanually.spec.js
+++ b/cypress/integration/old-bugs/request-saved-after-copymanually.spec.js
@@ -3,7 +3,7 @@ import { isOn, skipOn } from '@cypress/skip-test';
  * In #830, we almost changed the behaviour that requests should be saved after the modal opened without waiting until the modal closed.
  */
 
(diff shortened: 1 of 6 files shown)


src/Components/MailtoDropdown.js
```diff
@@ -37,19 +37,22 @@ export const MAILTO_HANDLERS = {
...
                 e.target.select();
                 e.target.focus();
                 previous_active_element = e.target;
             };
-            const modal = render(
+            createModal((state) => (
                 <IntlProvider scope="generator" definition={I18N_DEFINITION}>
                     <Modal
-                        positiveText={<Text id="ok" />}
+                        positiveText={t('ok', 'generator')}
```
baltpeter, line 55: Does that make a difference?

zner0L: Yes, this throws an error (in Chrome), because `<Text />` is an object, not a string.

src/Components/MailtoDropdown.js
```diff
@@ -37,19 +37,22 @@ export const MAILTO_HANDLERS = {
         countries: ['all'],
     },
     copymanually: {
-        onClick: (d) => {
-            const dismiss = () => render('', document.body, modal);
+        onClick: (d, createModal, onSuccess) => {
+            const dismiss = () => {
+                createModal(null);
+                if (onSuccess) onSuccess();
```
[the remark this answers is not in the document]
zner0L: There was a test for that, but I changed it (though it didn't test for it explicitly). I am not particularly invested in one solution or the other, the changed one just was easier because I was too lazy to properly manage the modal state (onSuccess might close the modal if the text modal is opened from another modal).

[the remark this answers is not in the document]
baltpeter: I'm I hadn't noticed this in the PR, I definitely would have fallen for that and expected clicking the "Copy text manually button" to save the request. And I'm pretty sure most users (well, those who understand the "My requests" workflow, anyway…) would as well.

> There was a test for that, but I changed it (though it didn't test for it explicitly).

Implicit tests don't count. As you just proved, someone will just change the test if it fails. :P

src/Components/MailtoDropdown.js
```diff
@@ -129,7 +128,9 @@ export default class MailtoDropdown extends Component {
                     if (!props.letter) e.preventDefault();
                     else {
                         if (MAILTO_HANDLERS[h].onClick)
-                            MAILTO_HANDLERS[h].onClick(data, props.createModal, props.onSuccess);
+                            MAILTO_HANDLERS[h].onClick(data, props.createModal).then(() => {
+                                props.onSuccess();
+                            });
```
baltpeter, line 133:
```suggestion
                            MAILTO_HANDLERS[h].onClick(data, props.createModal).then(props.onSuccess());
```

src/Components/RequestGeneratorBuilder.js
```diff
@@ -552,56 +552,55 @@ export default class RequestGeneratorBuilder extends Component {
...
+                            letter={state.letter}
+                            download_filename={state.download_filename}
+                            download_active={state.download_active}
+                            done={state.request.done}
+                            ready={state.ready}
+                            buttonText={t(
+                                state.request.transport_medium === 'email' ? 'send-email-first' : 'download-pdf-first',
+                                'generator'
+                            )}
+                            createModal={(modal) => new Promise((resolve, reject) => this.setState({ modal }, resolve))}
```
baltpeter, line 572:
```suggestion
                            createModal={(modal) => new Promise((resolve) => this.setState({ modal }, resolve))}
```

src/Components/RequestGeneratorBuilder.js
```diff
@@ -552,56 +552,55 @@ export default class RequestGeneratorBuilder extends Component {
     };
 
     confirmNewRequest = () => {
+        let confirmNewRequestModal = (state) => (
```
baltpeter, line 555:
```suggestion
        const confirmNewRequestModal = (state) => (
```

cypress/integration/old-bugs/request-saved-after-copymanually.spec.js
```diff
@@ -0,0 +1,52 @@
+import { isOn, skipOn } from '@cypress/skip-test';
+/*
+ * In #830, we almost changed the behaviour that requests should be saved after the modal opened without waiting until the modal closed.
+ */
+
+describe('Request should exists without clicking ok in the copy manually modal', () => {
```
baltpeter, line 6:
```suggestion
describe('Request should be saved without clicking ok in the copy manually modal', () => {
```

baltpeter (changes requested):
Other than the nits below, this is good to go.

src/Components/MailtoDropdown.js
zner0L: Isn't that wrong? I think you mean this:

```suggestion
                            MAILTO_HANDLERS[h].onClick(data, props.createModal).then(props.onSuccess);
```

baltpeter (approved):
Thanks, let's give this a go.

PR #831, opened 2021-12-04
Improve Hugo build speed by optimizing related company computation

baltpeter:
I felt like doing a bit of optimization for our Hugo build and the obvious place for that is the related companies template.

Through a series of relatively simple changes, I was able to get the build time from 27.7s down to 16.4s on my machine.


zner0L (approved):
This is really great. Sometimes the small changes make a big difference! This will probably even save us quite some money…

PR #835, opened 2021-12-24
Fix bugs introduced by #776

zner0L:


baltpeter (approved):
Thanks!

This PR introduces a new regression in that the title text for the "tested" and "imported/scraped" icons doesn't work anymore. But since that is a lot less bad than previously, I'm happy to merge this now and create an issue.

PR #851, opened 2022-02-13
add skip-to-content link, fixes #838

### LINKED ISSUE #838: Add "skip navigation" link

mal-tee:
I had to skip `stylelint`:

 > 113:4  ✖  Expected #skip-to-content is used with @media (prefers-reduced-motion)  a11y/media-prefers-reduced-motion
 
 Not sure why that triggers o.o

### DIFF
--- a/assets/styles/main.scss
+++ b/assets/styles/main.scss
@@ -112,7 +112,7 @@ a #logo {
 
 #skip-to-content {
     padding: 50px;
-    color: black;
+    color: $bg-color;
     background-color: $primary-color;
     position: absolute;
     transform: translateY(-200%);
--- a/assets/styles/main.scss
+++ b/assets/styles/main.scss
@@ -117,7 +117,7 @@ a #logo {
     position: absolute;
     transform: translateY(-200%);
     transition: transform 0.2s;
-    @media (prefers-reduced-motion: reduce) {
+    @media screen and (prefers-reduced-motion: reduce) {
         transition: transform 0s;
     }
     &:focus {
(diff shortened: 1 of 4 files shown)


baltpeter (changes requested):
Thanks, this is a good addition!

Two small comments:

* It doesn't work on the homepage, as there's no `#heading` there (I _think_ that's the one page where that's the case). We should probably special-case another element there. Another alternative would be to remove it for the homepage.
* I think it would still be obvious enough if we made the button blue instead of red.

assets/styles/main.scss
```diff
@@ -110,6 +110,21 @@ a #logo {
     }
 }
 
+#skip-to-content {
+    padding: 50px;
+    color: black;
```
baltpeter, line 115:
```suggestion
    color: $bg-color;
```

layouts/_default/baseof.html
```diff
@@ -44,6 +44,11 @@
 </head>
 
 <body>
+    {{ if .IsHome }}
+        <a id="skip-to-content" href="#heading">{{T "skip-to-content"}}</a>
+    {{ else }}
+        <a id="skip-to-content" href="#home-hero">{{T "skip-to-content"}}</a>
+    {{ end }}
```
baltpeter, line 51: I'd prefer:

```suggestion
    <a id="skip-to-content" href="{{ if .IsHome }}#home-hero{{ else }}#heading{{ end }}">{{T "skip-to-content"}}</a>
```

baltpeter (changes requested):

baltpeter (approved):
Thank you!

Issue #553, opened 2021-02-22
Fix unexpected behaviour of custom fields in ID data controls


baltpeter:
Not super happy with this implementation, unfortunately. I'm afraid, it causes some quite unexpected behaviour of fields vanishing, moving around and merging. And I am pretty sure that the last thing in the video is also a bug, though I am not sure what causes it:

![Peek 2020-11-14 19-51](https://user-images.githubusercontent.com/4048809/99154812-8f92f200-26aa-11eb-8811-96a503b6059a.gif)

_Originally posted by @baltpeter in https://github.com/datenanfragen/website/pull/485#discussion_r523450733_

PR #852, opened 2022-02-15
bug fix: convert string to regex

mal-tee:
I haven't tested this, but this really looks like a bug. :)


baltpeter:
@zner0L What do we want to do here?

The fix is obviously correct but the (intended) original implementation really isn't great. While this will now allow you to save a field called `name::fixed`, it instead causes `abc__de` and `abc::de` to collide. We should really just disallow `::` in the field name. Same for fields with the same name as an existing field.

If we merge this now, I think we should at least create a new issue to remind us to fix the underlying problem in the future. We do already have #553 but that is a separate problem.

zner0L:
This should really be solved by moving it all into a persisted zustand state now, as #892 calls for.

PR #859, opened 2022-02-28
Add Croatian translation

milotype:
Add Croatian translation with image


mal-tee:
Thank you!

Not sure why this fails to build.
 @baltpeter,  seems like it can't find the link target at the bottom, do we have to rename that to "my-requests"? I really don't know why it can't find the image tho.

baltpeter:
> @baltpeter, seems like it can't find the link target at the bottom, do we have to rename that to "my-requests"?

Exactly. `link` resolves the slug itself. You have to use the non-localized version (filename).

> I really don't know why it can't find the image tho.

Seems like something doesn't like non-ASCII filenames. :D

baltpeter (approved):
Thank you!

milotype:
@baltpeter Thank you for making the necessary corrections ...

PR #863, opened 2022-03-07
Add Croatian translation for contribute.me

milotype:


baltpeter (approved):
Thank you!

The test error is unrelated and will be fixed by https://github.com/datenanfragen/data/pull/1705, so we can ignore that here.

PR #872, opened 2022-03-18
TypeScriptify i18n functions and CommentsWidget

baltpeter:
See #868.


baltpeter:
I've narrowed the Cypress error down to this line:

zner0L (changes requested):
If you fix the tests and implement `useCallback`, this is fine.

baltpeter:
> If I remove that, it works fine. Setting an empty `paths` in Cypress' `tsconfig.json` doesn't help. Hm…

It genuinely seems to be caused by the `paths` property and not by the `globals.ts` that introduces. The error occurs even with an entirely empty `globals.ts`. Weird…

baltpeter:
Changing the filename from `globals.ts` to `globals.d.ts` seems to solve the problem. That is more correct anyway, as those files will only contain declarations, never implementations.

baltpeter:
Btw: I tend to make nice, scoped commits. Please don't squash them.

zner0L (approved):
Excellent. Love seeing the typing come together.

PR #873, opened 2022-03-31
Convert even more things to TypeScript

baltpeter:
Continuing the work for #868.


layouts/partials/supported-countries.html
```diff
@@ -1,21 +1,21 @@
...
-{{ $.Scratch.SetInMap "country_counts" . (add (index ($.Scratch.Get "country_counts") .) 1) }}
-{{ end }}
-{{ end }}
+    {{ range .Param "relevant-countries" }}
+        {{ if not (isset ($.Scratch.Get "country_counts") .) }}
+            {{ $.Scratch.SetInMap "country_counts" . 1 }}
+        {{ else }}
+            {{ $.Scratch.SetInMap "country_counts" . (add (index ($.Scratch.Get "country_counts") .) 1) }}
+        {{ end }}
+    {{ end }}
```
mal-tee, line 10: I wish there was a stylelint-esque tool for these templates. :D

baltpeter: Oh yeah, I'd be happy even just with IDE support for them. :D

src/Components/Radio.tsx
```diff
@@ -0,0 +1,25 @@
+import type { JSX } from 'preact';
+
+type RadioProps = {
+    id: string;
+    label: string;
+    radioVariable: string;
+    value: string;
+    name: string;
+    onChange: JSX.EventHandler<JSX.TargetedEvent<HTMLInputElement, Event>>;
```
baltpeter, line 9: @zner0L I'd like these to be passed `e.target.value` directly, the DOM event seems like an irrelevant implementation detail (and it needs annoying TS casting on the using side). But that would require quite extensive changes, so I'll hold off on that for now.

src/Components/Modal.tsx
```diff
@@ -0,0 +1,106 @@
...
+    const options: typeof default_options & ModalOptions = useMemo(
+        () => ({ ...default_options, ..._options }),
+        [_options]
+    );
+
+    const [shown, setShown] = useState(options.shownInitially);
+    if (options.shownInitially) options.onShown?.();
+
+    const show = useCallback(() => setShown(true), []);
+    const dismiss = useCallback(() => setShown(false), []);
```
zner0L, line 42: These should depend on `setShown`, shouldn't they?

src/Components/Modal.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+        const listener = (e: KeyboardEvent) => e.key === 'Escape' && dismiss();
+        document.body.addEventListener('keydown', listener);
+
+        return () => document.body.removeEventListener('keydown', listener);
+    }, [dismiss, shown, options.escDismisses]);
+
+    useEffect(() => {
+        if (shown) options.onShown?.();
+        else options.onDismissed?.();
```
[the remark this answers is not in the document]
zner0L: Yup, just confirmed this.

src/Components/I18nWidget.tsx
```diff
@@ -0,0 +1,121 @@
...
+};
+
+export const I18nWidget = (props: I18nWidgetProps) => {
+    // Don't ever update `country` directly but rather use `globals.country`.
+    const [country, setCountry] = useState(window.globals.country);
+    window.globals._country_listeners.push((new_country) => setCountry(new_country));
+
+    const [newLanguage, setNewLanguage] = useState(window.LOCALE);
+
+    const [Modal, showModal, dismissModal] = useModal(<MarkupText id="language-change-modal" />, {
```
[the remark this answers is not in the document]
zner0L: After looking at it some more, the problem is not the modal, but that you don't allow to reset `newLanguage`, so the modal will only be shown once (c.f. `changeLanguage`).

src/Components/MailtoDropdown.tsx
```diff
@@ -0,0 +1,171 @@
...
+        ),
+        body: encodeURIComponent(props.letter.toEmailString()),
+    };
+
+    const onCopyManuallyInputClick = (e: JSX.TargetedMouseEvent<HTMLInputElement | HTMLTextAreaElement>) => {
+        if (previous_active_element.id === e.currentTarget.id) return;
+
+        e.currentTarget.select();
+        e.currentTarget.focus();
+        previous_active_element = e.currentTarget;
```
zner0L, line 72: Why is this not a state?

src/Components/MailtoDropdown.tsx
```diff
@@ -0,0 +1,171 @@
...
+                    id="mailto-dropdown-copymanually-body"
+                    className="form-element"
+                    rows={10}
+                    onClick={onCopyManuallyInputClick}
+                    readOnly>
+                    {decodeURIComponent(data.body)}
+                </textarea>
+            </div>
+        </IntlProvider>,
+        { onPositiveFeedback: () => dismissCopyManuallyModal(), positiveText: t('ok', 'generator') }
```
zner0L, line 114: `previous_active_element` should be reset on dismiss!

src/Components/SearchBar.tsx
```diff
@@ -0,0 +1,220 @@
...
+        emptyTemplate,
+        filters,
+        footerTemplate,
+        headerTemplate,
+        index,
+        numberOfHits,
+        onAutocompleteSelected,
+        queryBy,
+        suggestionTemplate,
+    } = props;
```
zner0L, line 76: You could also just use that as the argument.

src/my-requests.tsx
```diff
@@ -0,0 +1,342 @@
...
+    const clearRequests = useCallback(() => {
+        if (window.confirm(t('modal-clear-requests', 'privacy-controls'))) {
+            user_requests
+                .clearRequests()
+                .then(() => setRequests({}))
+                .catch((error) => rethrow(error, 'Could not clear requests.'));
+        }
+    }, []);
+
+    if (!Privacy.isAllowed(PRIVACY_ACTIONS.SAVE_MY_REQUESTS)) {
```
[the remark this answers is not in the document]
baltpeter: That's as far up as I can put it: https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

src/my-requests.tsx
```diff
@@ -0,0 +1,342 @@
...
+
+const RequestList = () => {
+    const [requests, setRequests] = useState<Record<string, UserRequest>>({});
+    const [selectedRequestIds, setSelectedRequestIds] = useState<string[]>([]);
+
+    useEffect(() => {
+        user_requests.getRequests().then((new_requests) => setRequests(new_requests));
+    }, []);
+
+    const sortedRequestIds = useMemo(
```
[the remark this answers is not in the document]
baltpeter: As are plain objects but at least from a quick glance, both are sorted in insertion order and can't have a custom sort function.

src/Components/Footnote.tsx
```diff
@@ -1,15 +1,22 @@
...
 
     useEffect(() => {
-        //Adapted after: https://usehooks.com/useOnClickOutside/
-        const listener = (event) => {
-            if (detailsRef.current?.open && !detailsRef.current?.contains(event.target)) {
-                detailsRef.current.removeAttribute('open');
+        // Adapted after: https://usehooks.com/useOnClickOutside/
+        const listener = (event: MouseEvent | TouchEvent) => {
+            const details_el = detailsRef.current as HTMLDetailsElement | null;
+            if (details_el?.open && !details_el?.contains(event.target as Element)) {
```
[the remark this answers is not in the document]
baltpeter: No, that would break it. This is one of the few cases where we explicitly need whatever element caused the event, not the one that has the event handler.

src/Components/I18nWidget.tsx
```diff
@@ -0,0 +1,130 @@
...
+    );
+    alreadyShowingModal = true;
+};
+
+export const I18nWidget = (props: I18nWidgetProps) => {
+    // Don't ever update `country` directly but rather use `globals.country`.
+    const [country, setCountry] = useState(window.globals.country);
+    window.globals._country_listeners.push((new_country) => setCountry(new_country));
+
+    const language_options = (Object.keys(window.SUPPORTED_LANGUAGES) as (keyof typeof window.SUPPORTED_LANGUAGES)[])
```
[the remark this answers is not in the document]
baltpeter: That's one of those "TS knows more about JS than you" cases: https://fettblog.eu/typescript-better-object-keys/#further-reading

src/Components/SvaFinder.tsx
```diff
@@ -141,119 +157,116 @@ const steps = deepmerge(
...
-            const sorted_keys = Object.keys(entries).sort((a, b) => {
-                // For the countries, move the user's country to the top of the list.
-                if (steps.country[globals.country]) {
-                    if (a === globals.country) return -1;
-                    else if (b === globals.country) return 1;
-                }
+    let content: JSX.Element;
+    if (state.result) {
+        if (props.callback) {
+            fetchSvaDataBySlug(state.result).then((sva) => props.callback?.(sva));
```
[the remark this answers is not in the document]
baltpeter: > Is this valid syntax?

The `props.callback?.(sva)`? Yes: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#optional_chaining_with_function_calls

src/Components/SearchBar.tsx
zner0L: Really? It doesn't complain to me. To be clear, what I am suggesting looks like this:

```typescript
const RealSearchBar = ({
    id,
    anchorize,
    debug,
    disableCountryFiltering,
    emptyTemplate,
    filters,
    footerTemplate,
    headerTemplate,
    index,
    numberOfHits,
    onAutocompleteSelected,
    queryBy,
    suggestionTemplate,
    placeholder,
    style,
}: SearchBarProps) => {
```

baltpeter: Oh okay, I misunderstood. Yeah, that works of course. I don't really have a particular preference either way.

src/DataType/UserRequests.ts
```diff
@@ -0,0 +1,56 @@
...
+    type: 'access' | 'erasure' | 'rectification' | 'objection' | 'custom';
+    response_type?: 'admonition' | 'complaint';
+    slug: string;
+    recipient: string;
+    email: string;
+    via: 'fax' | 'letter' | 'email';
+};
+
+export class UserRequests {
+    storeRequest(db_id: string, item: UserRequest) {
```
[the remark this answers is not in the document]
baltpeter: Oh, you're right of course (how'd I miss that? \*shakeshead*). But I feel like checking that should definitely be the responsibility of `UserRequests`, not its users…

src/DataType/UserRequests.ts
```diff
@@ -0,0 +1,56 @@
...
+    getRequest(db_id: string) {
+        return localforage_instance.getItem(db_id);
+    }
+
+    getRequests() {
+        const requests: Record<string, UserRequest> = {};
+        return new Promise<Record<string, UserRequest>>((resolve, reject) => {
+            localforage_instance
+                .iterate<UserRequest, void>((data, reference) => {
+                    requests[reference] = data;
```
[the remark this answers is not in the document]
baltpeter: I'd say in the interest of getting this merged quickly, this should definitely be moved into an issue. :D

src/my-requests.tsx
```diff
@@ -0,0 +1,342 @@
...
+                            <a
+                                className="button button-primary"
+                                href={`${window.BASE_URL}generator`}
+                                style="float: right;">
+                                <Text id="generate-request" />
+                            </a>
+                            <div className="clearfix" />
+                        </div>
+                    ) : (
+                        [
```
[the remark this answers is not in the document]
baltpeter: Any benefit to that?

src/Components/FlashMessage.tsx
```diff
@@ -0,0 +1,48 @@
...
+                className="button-unstyled close-button icon-close"
+                onClick={dismiss}
+                title={t('cancel', 'generator')}
+            />
+            <div className="inner">{props.children}</div>
+        </div>
+    );
+};
+
+const flash_messages: VNode<FlashMessageProps>[] = [];
```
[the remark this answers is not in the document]
baltpeter: I guess…

src/Components/I18nWidget.tsx
```diff
@@ -0,0 +1,130 @@
...
+                )}
+            </div>
+        </IntlProvider>
+    );
+};
+
+export const I18nButton = () => {
+    // Don't ever update `country` directly but rather use `globals.country`.
+    const [country, setCountry] = useState(window.globals.country);
+    window.globals._country_listeners.push((new_country) => setCountry(new_country));
```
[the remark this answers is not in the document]
baltpeter: Well, my next commit would have been getting rid of `globals` and replacing it with a zustand store. But if you depend on the changes in this PR, I'd say we merge this as is and then I make a new one for those changes.

src/Components/SvaFinder.tsx
```diff
@@ -141,119 +157,116 @@ const steps = deepmerge(
...
-        let content;
+export const SvaFinder = (props: SvaFinderProps) => {
+    const [state, setState] = useState<SvaFinderState>(initial_state);
 
-        if (this.state.result) {
-            if (typeof this.props.callback === 'function') {
-                fetchSvaDataBySlug(this.state.result).then((sva) => this.props.callback(sva));
-                return <p>{t('loading-sva', 'sva-finder')}</p>;
-            }
+    const selectOption = useCallback(
```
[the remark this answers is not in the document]
baltpeter: If you say so. \*shrug*

src/Components/SvaFinder.tsx
```diff
@@ -141,119 +157,116 @@ const steps = deepmerge(
...
+        }, {});
+        const sorted_keys = Object.keys(entries).sort((a, b) => {
+            // For the countries, move the user's country to the top of the list.
+            if (a === window.globals.country) return -1;
+            else if (b === window.globals.country) return 1;
 
-                // Otherwise, just sort alphabetically.
-                return entries[a].localeCompare(entries[b]);
-            });
+            // In the first step for Germany, "Any other public or private entity" has to be sorted last.
```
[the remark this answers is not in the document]
baltpeter: Um, it is for me and I definitely remember adjusting the height to ensure that it is visible.

image

src/Components/SvaFinder.tsx
```diff
@@ -141,119 +157,116 @@ const steps = deepmerge(
...
-    };
-}
+    return (
+        <div className="sva-finder box box-info" style={props.style}>
+            {props.callback || win.props?.showTitle === false ? '' : <h2>{t('sva-finder', 'sva-finder')}</h2>}
+
+            {content}
+
+            <div style="margin-top: 20px;">
+                {state.prev_state && (
```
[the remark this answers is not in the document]
baltpeter: Nope. I think TS is trying to protect us from a TOCTOU here. But I don't care. :D

src/Components/Modal.tsx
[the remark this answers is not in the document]
baltpeter: Ugh, yeah. That's annoying. I don't really feel like carrying extra state around for something we aren't using anyway. So bye bye, `onShown` and `onDismissed` \*wave*. You'll be (dis)missed.

src/Components/MailtoDropdown.tsx
baltpeter: I remember that I had a reason for that but I don't remember what it was. :( You could watch the video of me implementing the MailtoDropdown to find out. :P

src/Components/MailtoDropdown.tsx
baltpeter: I know my fix doesn't solve the problem when you just click outside the modal. But _please_ don't make me bring back `onDismissed` just for that. .D

src/DataType/UserRequests.ts
```diff
@@ -0,0 +1,66 @@
...
+    email: string;
+    via: 'fax' | 'letter' | 'email';
+};
+
+export class UserRequests {
+    localforage_instance?: typeof localforage;
+
+    constructor() {
+        if (Privacy.isAllowed(PRIVACY_ACTIONS.SAVE_MY_REQUESTS)) {
+            this.localforage_instance = localforage.createInstance({
```
zner0L, line 21: What if the constructor is called, if `Privacy.isAllowed(PRIVACY_ACTIONS.SAVE_MY_REQUESTS)` is false, but the state later changes and there is no `localforage_instance` available?

src/Components/SvaFinder.tsx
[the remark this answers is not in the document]
zner0L: Why not just sort it to the top? We are trying to optimise for mobile here, aren't we?

src/Components/MailtoDropdown.tsx
zner0L: You didn't have a good reason. You can even just move the id into the state instead of the whole element.

src/Components/Modal.tsx
zner0L: Let's put `onDismiss`and `onShown` back here.

zner0L (changes requested):
There are some issues remaining, but as far as I am concerned, this can be merged and the changes addressed later.

zner0L:
@baltpeter Please take another look at my changes now. There is a problem with `UserRequests` because the changes modify the expected behavior in the legacy code, e.g. the "Clear Requests" button in the modal on the Privacy page won't work. I think we need to make a decision here, how we want to move forward.

src/DataType/UserRequests.ts
```diff
@@ -14,36 +14,39 @@ export type UserRequest = {
 };
 
 export class UserRequests {
-    localforage_instance?: typeof localforage;
+    __localforage_instance?: typeof localforage;
```
baltpeter, line 17: JS does have proper private fields: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

src/DataType/UserRequests.ts
```diff
@@ -14,36 +14,39 @@ export type UserRequest = {
...
 
-    constructor() {
+    getLocalforageInstance() {
```
baltpeter, line 19: Any reason this isn't a [proper getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get)? 

That way, we could keep the existing `this.localforage_instance` accesses as-is.

baltpeter:
> 

It is really unfortunate that localForage doesn't make it possible to straight-up delete a store. But I just noticed that clearing a non-existent store doesn't create an empty one. This seems unexpected to me, so maybe you should try confirming that as well in case I missed something. But if that is the case, clearing should definitely work regardless of the privacy controls.

src/Components/Modal.tsx
```diff
@@ -34,8 +36,15 @@ export const useModal = (children: ComponentChildren, _options?: ModalOptions) =
 
     const [shown, setShown] = useState(options.shownInitially);
 
-    const show = useCallback(() => setShown(true), []);
-    const dismiss = useCallback(() => setShown(false), []);
+    const show = useCallback(() => {
+        setShown(true);
+        options.onShow?.();
+    }, [options.onShow]);
```
baltpeter, line 42: ESLint isn't happy about that:

image

As far as I understand it, you either need to depend on `options` as a whole or destructure it before.

src/DataType/UserRequests.ts
zner0L: Forgot about those! Damn you, python.

src/DataType/UserRequests.ts
zner0L: Typescript wasn't happy about this.

zner0L: I think there was a warning about the target being to low.

src/Components/Modal.tsx
zner0L: Yes, but in the same file it accepts similar syntax for another property. I think this should be right, but maybe I shouldn't doubt ESlint.

zner0L (commented):
This looks good now. Thanks for the big effort!

zner0L (approved):

mal-tee:
wohoo

PR #885, opened 2022-05-06
Move window.globals into a zustand store

baltpeter:
Parts of the implementation are unfortunately a bit hacky. I'm open to better suggestions but we need to work with quite annoying constraints. Luckily, the hackiness at least is entirely limited to the store's implementation, actually using that (i.e. the important part) is pleasant.


src/Utility/browser.ts
```diff
@@ -89,3 +91,17 @@ export const clientPost = (
     form.submit();
     document.body.removeChild(form);
 };
+
+// This uses the `navigator.language` property (similar-ish to the `Accept-Language` header which we cannot access from
+// JS) which may not necessarily represent the user's country (or even include region information at all).
+// The more reliable way would be to feed the user's IP into a geolocation service but that is not an option, so we have
+// to stick with this.
+export const guessUserCountry = (): Country => {
+    const { country } = parseBcp47Tag(navigator.language) as { country: Country };
```
zner0L, line 100: The type annotation should really be a proper return type annotation on the function…

src/Utility/browser.ts
```diff
@@ -89,3 +91,17 @@ export const clientPost = (
...
+
+    // If we cannot guess the user's country, it makes sense to fallback to the language.
+    if (!navigator.language || !country) return fallback_countries[window.LOCALE];
```
zner0L, line 103: Can we assume `window.LOCALE` to always be well defined? Should we harden our code for that? Depending on that, I would expect this to be undefined in some cases.

src/Components/SvaFinder.tsx
```diff
@@ -168,6 +169,8 @@ const initial_state: SvaFinderState = {
 };
 
 export const SvaFinder = (props: SvaFinderProps) => {
+    const { country } = useAppStore();
```
zner0L, line 172:
```suggestion
    const country = useAppStore((state) => state.country);
```

src/Components/SearchBar.tsx
```diff
@@ -71,6 +72,8 @@ const RealSearchBar = ({
     placeholder,
     style,
 }: SearchBarProps) => {
+    const { country } = useAppStore();
```
zner0L, line 75:
```suggestion
    const country = useAppStore((state) => state.country);
```

src/Components/MailtoDropdown.tsx
```diff
@@ -50,6 +51,8 @@ const mailto_handlers = createMailtoHandlers({
 });
 
 export const MailtoDropdown = (props: MailtoDropdownProps) => {
+    const { country } = useAppStore();
```
zner0L, line 54:
```suggestion
    const country = useAppStore((state) => state.country);
```

src/Components/MailtoDropdown.tsx
```diff
@@ -130,7 +133,7 @@ export const MailtoDropdown = (props: MailtoDropdownProps) => {
     const handlers =
```
zner0L, line 133: Maybe it is smart to memoize these.

src/Components/I18nWidget.tsx
```diff
@@ -10,9 +11,7 @@ type I18nWidgetProps = {
 };
 
 export const I18nWidget = (props: I18nWidgetProps) => {
-    // Don't ever update `country` directly but rather use `globals.country`.
-    const [country, setCountry] = useState(window.globals.country);
-    window.globals._country_listeners.push((new_country) => setCountry(new_country));
+    const { country, changeCountry } = useAppStore();
```
zner0L, line 14:
```suggestion
    const country = useAppStore((state) => state.country);
    const changeCountry = useAppStore((state) => state.changeCountry);
```

src/Components/I18nWidget.tsx
```diff
@@ -80,9 +79,7 @@ export const I18nWidget = (props: I18nWidgetProps) => {
                         <div className="select-container">
                             <select
                                 value={country}
-                                onChange={(e) => {
-                                    window.globals.country = e.currentTarget.value as typeof window.globals.country;
-                                }}>
+                                onChange={(e) => changeCountry(e.currentTarget.value as typeof country)}>
```
zner0L, line 82: This should probably also be the `Country` type directly, for consistency.

src/hooks/useFetch.ts
```diff
@@ -0,0 +1,72 @@
...
+                return { ...initialState, error: action.payload };
+            default:
+                return state;
+        }
+    };
+
+    const [state, dispatch] = useReducer(fetchReducer, initialState);
+
+    useEffect(() => {
+        if (!url) return;
```
zner0L, line 36: Why is `url` optional then?

src/hooks/useFetch.ts
```diff
@@ -0,0 +1,72 @@
...
+
+        cancelRequest.current = false;
+
+        const fetchData = async () => {
+            dispatch({ type: 'loading' });
+
+            if (cache.current[url]) {
```
zner0L, line 43: Maybe we should document somewhere that this fetch can not be used to fetch rapidly changing content, as all subsequent requests will be served from the cache…

src/Components/Wizard.tsx
```diff
@@ -21,13 +23,13 @@ const categories = [
 
 const saved_companies = Privacy.isAllowed(PRIVACY_ACTIONS.SAVE_WIZARD_ENTRIES) ? new SavedCompanies() : undefined;
 
+type CompanyNameToSlug = Record<string, string>;
+
 export const Wizard = () => {
-    // Don't ever update `country` directly but rather use `globals.country`.
-    const [country, setCountry] = useState(window.globals.country);
-    window.globals._country_listeners.push((new_country) => setCountry(new_country));
+    const { country } = useAppStore();
```
zner0L, line 29:
```suggestion
    const country = useAppStore((state) => state.country);
```

zner0L (changes requested):
This is just a bit hacky, but I agree that it is fine. I am a little worried about persist creating a store we might not want. This is not an issue now, as we also always set the country cookie, but it will be, if we start storing other stuff in there. Other than that, I can only see some minor problems,

baltpeter:
> I am a little worried about persist creating a store we might not want. This is not an issue now, as we also always set the country cookie, but it will be, if we start storing other stuff in there.

As localstorage doesn't have multiple DBs like IndexedDB, the entire store is stored in a single JSON blob, so I don't think this matters too much.

And even then, it's easy enough to override the store function to discard the value if it is being set to the default.

cypress/integration/old-bugs/donation-widget-nan.js
```diff
@@ -5,7 +5,6 @@
 
 describe('Donation widget amount', () => {
     beforeEach(() => {
-        cy.setCookie('country', 'en'); // to make sure that the language suggestion message doesn't interfere
```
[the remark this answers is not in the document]
baltpeter: I don't see how the modal would interfere.

src/Utility/browser.ts
baltpeter: But it isn't guaranteed that `parseBcp47Tag()` always results in something that satisfies our `Country` type. I think it's better to be aware of that in downstream usages and cast there when we consider it appropriate.

src/Components/MailtoDropdown.tsx
baltpeter: I mean, we can… But I really struggle to decide when that's a good idea, especially since you barely used `useCallback()` in your PR at all, even for stuff where I definitely would have. :D 

In this case, computing `handlers` requires ~~ten~~five simple equality checks on strings. I wouldn't be surprised if the memoization was more complicated than that.

src/hooks/useFetch.ts
baltpeter: I guess it might be useful to be able to do `useFetch(var_that_might_be_undefined)`?

src/hooks/useFetch.ts
```diff
@@ -0,0 +1,77 @@
...
+};
+
+interface State<T> {
+    data?: T;
+    error?: Error;
+}
+type Cache<T> = { [url: string]: T };
+type Action<T> = { type: 'loading' } | { type: 'fetched'; payload: T } | { type: 'error'; payload: Error };
+
+export const useFetch = <T = unknown>(url?: string, options?: Options, fetch_options?: RequestInit): State<T> => {
```
zner0L, line 16: I think the order of parameters should be the other way around. It is more likely to leave the `useFetch` specific options empty and that would be annoying.

```suggestion
export const useFetch = <T = unknown>(url?: string, fetch_options?: RequestInit, options?: Options): State<T> => {
```

src/Utility/browser.ts
zner0L: Yes, on second thought I realized that this falls out of hugo…

src/Utility/browser.ts
zner0L: I am not completely sold on this… Firstly, the cast here should definitely include `undefined` (I know this changes nothing, but casting a potentially undefined value to `Country` is wrong and could lead to bugs in the future). Secondly, for this function to be safe, this requires proper type checking and maybe a typeguard. I can see how we do not need in in our case (although I think the more elegant solution would be to define `window.SUPPORTED_COUNTRIES.includes(country)` as a typeguard), but in other cases it might. I can see how this is something that can happen downstream.

src/general.tsx
```diff
@@ -60,30 +50,16 @@ function notifyOtherLanguages(preferred_language: string, website_language: stri
...
+    Cookie.remove('country');
 
-    // If we cannot guess the user's country, it makes sense to fallback to the language.
-    if (!navigator_lang || !bcp47_country) return fallback_countries[window.LOCALE];
+    const { language: preferred_language } = parseBcp47Tag(navigator.language);
+    const { language: website_language } = parseBcp47Tag(document.documentElement.lang);
 
-    // If however we *can* guess the country but just don't support it, we show all companies.
-    return window.SUPPORTED_COUNTRIES.includes(bcp47_country) ? bcp47_country : 'all';
+    if (preferred_language !== website_language && preferred_language! in fallback_countries)
```
zner0L, line 61: This is probably just a staging bug, but If I change languages in the netlify build preview, I am not shown the choose language modal.

zner0L (changes requested):
Just a few additional comments, should be quick to address.

src/general.tsx
baltpeter: Um, I can't reproduce that.

tFxNwRv8cr

zner0L: Ah, no. I am talking about the language suggestion flash.

baltpeter: Well yeah. If you change the language on the staging site, you're still on the same origin, so your country setting (which determines whether the suggestion is shown) is preserved.

src/hooks/useFetch.ts
zner0L: I feel like your new implementation is a little clumsy… Don't we use the `fetchOptions` on nearly every request?

baltpeter: Do we? I just looked through all our uses of `fetch()`. Of the ones where it would make sense to replace them with this, none set any options.

There are only three instances where we set options but those are all `POST`/`PUT` and I definitely wouldn't use `useFetch` for that.

zner0L (approved):
Looks good now. Thanks for clearing up the open questions!

PR #887, opened 2022-05-10
More TS yet (for #868)

baltpeter:


src/Components/Accordion.tsx
```diff
@@ -0,0 +1,35 @@
+import type { ComponentChildren } from 'preact';
+import { useState } from 'preact/hooks';
+
+type AccordionProps = {
+    id: string;
+    title: string;
+    expanded: boolean;
```
baltpeter, line 7: We should rename that to `expandedInitially` or something like that.

zner0L: Yes, please do that.

zner0L (dismissed):
If you don't plan on implementing your own suggestion, this is ready to be merged.

baltpeter:
We'll have to remember the rename when merging #874.

zner0L (approved):

PR #895, opened 2022-05-31
New UX design for the request generator

baltpeter:


baltpeter (commented):
My notes from a quick look over the new design. If you want, I'll happily take over and implement those myself if you want to do something else for a change (feel free to pick up the WASM thing for example).

* [ ] Company packs: "Show more"/"Show less"
* [ ] Company packs can have descriptions.
* [ ] Not too keen on the company packs design. Do we need an `<ul>` there?
* [ ] The company packs are done. Merge and integrate that before the user testing.
* [ ] I know I've said it before but I don't think the blue background is an intuitive way to indicate a selected state. This seems more like a hovered/focussed state. This is especially true if there is only one result (e.g. if you search for "google").
* [x] Centered next button seems unconventional. I'd put it on the right to indicate the "forward" motion.
* [x] The "review selection" page needs an explanatory text. It's unclear what the user is supposed to do there.
* [x] The `runs` entries need to be selected by default.
* [x] I don't like the implementation of the ID data description inputs. The jumping around feels wrong. I was thinking something like this: https://codesandbox.io/s/exciting-tdd-qm1re7
* [ ] The spacing between the descriptions and value inputs is too big. Put the dynamic input container into a centered `.col80` or whatever.
* [ ] Generating the A1 request says "[object Object]" at the top.
* [x] The text in the (previously) "copy manually" dialog is wrong for the new use case.
* [x] We decided to only enable the "next request" button once the request has been sent.
* [x] In the "send request" step, the heading shouldn't just be the company name but "Request to <c_name>" (but better).
* [x] The "add input" button feels _really_ out of place there. It's completely misaligned!
* [ ] Not too crucial but the "add input" drop…up is cut off if it has too many elements.
* [ ] We need to get rid of the flash of the "send request" modal.
* [x] Signature input is missing.
* [x] We wanted to have a "skip company" button.
* [ ] This doesn't have to be implemented for the first UX testing, but I we wanted to persist the state.
* [ ] Not important: In the arrow thing at the top, long company names should be elided instead of spanning multiple rows.

zner0L:
Well, I agree with most of these things, they just weren't a priority. *Please* iterate on the company packs design, I hate it. a couple of things I disagree with:
- I want to keep the blue selected sate and I think we should take this into user testing to see whether this is an actual problem. Ich think checkboxes are ugly.
- `runs` entries are selected by default, I was just too lazy to do that in the fixture data as well.
- I get what you want for the description inputs, but keeping the buttons is an a11y feature IMO. That is why I didn't implement it in you suggested way (I also wouldn't know where to put the delete button).
- I like the spacing, but if you disagree, please remember to remove the spacing on mobile.
- The `[object Object]` is a bug in `letter-generator`, your are right, I will fix that.
- The "Add input" button is centered. I think it mainly looks weird because of the border you wanted.

baltpeter:
My notes from a user test:

- [ ] Not obvious enough that you can scroll in company pack list.
- [ ] Buttons sometimes jump around:
	- [ ] Sometimes when removing company on "review selection" page.
	- [ ] When toggling the "include objection" checkbox for erasure requests.
- [ ] If user closes send modal on "fill requests" page, it isn't obvious how to advance to the next request (the "next" button is only in the modal).
- [ ] Check whether all generated requests are recorded in "my requests".
- [ ] The page flash when the next company is being fetched in "fill requests" is unfortunate. Maybe preload that?
- [ ] The "send a request" button on company single pages needs to link to the new app now and prefill that. This should enable a simplified workflow with fewer steps. We know the user only wants to send one request to this particular company. => No need to ask for more companies and selection review). If they _do_ want to send more requests, they can easily do that from the "what's next" page.
- [ ] Do we need to show the "review selection" page if the user only selected a single company? (-> But this would make it impossible to choose the runs entries in that case and might surprise users?)
- [x] Counts in breadcrumbs are sometimes wrong. (-> Was probably caused by my "skipped" changes, I've changed that to instead remove the company from the batch now, which should solve the problem.)
- [ ] Italic text is search highlights is not perceivable enough.
- [ ] Shouldn't keyboard navigation (arrow keys) already be possible when the search bar is still focused? This was the case in the autocomplete.
- [ ] Also accept Enter (instead of just Space) on a focused search entry? Should that have different behaviour?
- [ ] It's not clear enough that the user should change the field description when adding a new custom field. Maybe highlight that somehow?

src/i18n/en.json
```diff
@@ -181,7 +181,8 @@
         "footer-social-github-desc": "Find us on GitHub.",
         "footer-social-matrix-desc": "Talk with us through Matrix.",
         "footer-social-mastodon-desc": "Follow us on Mastodon.",
-        "footer-social-twitter-desc": "Follow us on Twitter."
+        "footer-social-twitter-desc": "Follow us on Twitter.",
+        "noscript-app": "The app runs only on your computer and is therefore completely written in JavaScript. The code won't send any data to any servers if you do not consent. If you want to use the app, please activate JavaScript in your browser or add an exception to your blocker. If it still doesn't work after that, send us an email at <a href='mailto:dev@datenanfragen.de'>dev@datenanfragen.de</a>."
```
baltpeter, line 185:
```suggestion
        "noscript-app": "The app runs only on your computer and is therefore completely written in JavaScript. The code won’t send any data to any servers if you do not consent. If you want to use the app, please activate JavaScript in your browser or add an exception to your blocker. If it still doesn’t work after that, send us an email at <a href='mailto:dev@datenanfragen.de'>dev@datenanfragen.de</a>."
```

src/types/company.d.ts
```diff
@@ -7,6 +7,8 @@ type CommonSchema = {
     slug: string;
     name: string;
     runs?: string[];
+    /** We copy the runs in this field to allow users to create thei own selection of companies in the orginal runs field. Leave this undefined to default to the standard `runs` field. Make this an empty array to unselect all runs entities. */
```
baltpeter, line 10:
```suggestion
    /// We copy the runs in this field to allow users to create their own selection of companies in the original runs field. Leave this undefined to default to the standard `runs` field. Make this an empty array to unselect all runs entities.
```

baltpeter (commented):
I'm fine with merging this into `new_ux` if the comments below are addressed. I've moved everything else to https://github.com/datenanfragen/website/issues/910, that can be addressed later.

@zner0L You should also do a pass over my changes.

src/Components/App/CompanySearchPage.tsx
```diff
@@ -94,55 +107,105 @@ const Results = connectStateResults(
...
-                        </button>
-                    </li>
+        <IntlProvider scope="company-packs" definition={window.I18N_DEFINITION}>
+            {/* TODO: This breaks the flex layout. -.- */}
+            <Modal />
+
+            <section className="company-suggestion-pack">
+                <h1>{t(`${pack.slug as 'address-brokers'}-title`, 'company-packs')}</h1>
+                {pack.type === 'add-all' && (
+                    <span className="icon icon-star color-yellow-600" title={t('add-all-star', 'company-packs')} />
```
zner0L, line 186: I think the yellow color is ugly.

src/Components/App/CompanySearchPage.tsx
```diff
@@ -94,55 +107,105 @@ const Results = connectStateResults(
...
+                        .slice(0, 7)
+                        .map((c) => c.name)
+                        .join(', ')}
+                    {pack.companies.length > 7 && (
+                        <em>
+                            <Text id="and-n-more" fields={{ n: pack.companies.length - 7 }} />
+                        </em>
+                    )}
+                </p>
+                <button className="company-suggestion-pack-cta button" onClick={showModal}>
```
zner0L, line 199: I think we need to indicate somehow if (some of) a company pack was added. Currently, the user can not be sure just from the company packs UI.

src/Components/App/SendRequestButton.tsx
```diff
@@ -2,23 +2,29 @@ import { useCallback, useRef, useEffect } from 'preact/hooks';
...
+    const [batch, resetRequestToDefault, markCurrentBatchCompanyDone, markCurrentBatchCompanySkipped] =
+        useGeneratorStore((state) => [
+            state.batch,
             state.resetRequestToDefault,
             state.markCurrentBatchCompanyDone,
             state.markCurrentBatchCompanySkipped,
-        ]
-    );
+        ]);
+    const remainingBatchEntries = Object.values(batch || {}).filter((e) => !e.done && !e.skipped).length || 0;
```
zner0L: This variable name is a little confusing. Maybe add `Number` to the end?

zner0L (approved):
Seems great as far as I can see. Looking forward to finally merging this.

assets/styles/generator.scss
```diff
@@ -59,7 +59,7 @@ textarea:placeholder-shown + label.fancy-label {
 /* Inputs */
 .dynamic-input-controls {
     display: flex;
-    justify-content: center;
```
zner0L: This breaks the dropup:
Screen Shot 2022-07-15 at 12 33 41

zner0L (changes requested):

PR #899, opened 2022-05-31
Prettier: Use prettier-plugin-organize-imports

baltpeter:
As we discussed, inconsistent import ordering and having to think about stuff like that are way too annoying. In #898, I didn't agree with where you put the type imports and that finally pushed me over the edge and I went looking for a solution. :D 

I personally would have sorted the imports differently than this plugin but it uses the official TypeScript language service API and I **much** prefer not having to think about the imports at all.

If we actually do care about controlling the sort order (I don't think I do), there's also https://github.com/trivago/prettier-plugin-sort-imports.

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,6 +113,7 @@
         "postcss-preset-env": "^7.0.1",
         "postcss-url": "^10.1.3",
         "prettier": "^2.5.0",
+        "prettier-plugin-organize-imports": "^2.3.4",
         "speed-measure-webpack-plugin": "^1.5.0",
         "stylelint": "^14.1.0",
         "stylelint-a11y": "^1.2.1",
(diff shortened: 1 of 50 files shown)


zner0L:
The way I see it, this is clearly a bug. None of the imports were supposed to be deleted, but it deletes them anyhow. I do have "format on save" enabled, but I also want to keep it that way. An yes, ESLint already takes care of unused imports. If you care a lot about those, we can compromise on increasing the severity of unused imports in ESLint so that the commit test fails for them. But this plugin is counterintuitve and, frankly, buggy.

baltpeter:
Note to self: https://github.com/simonhaenisch/prettier-plugin-organize-imports/issues/37#issuecomment-1217523372

christian-weiss:
This PR is 1.5 years old and 0.5 years stale. I vote for it to be CLOSED.

You may want to experiment with latest version (4.3.0), as this PR was done with 2.3.4

PR #902, opened 2022-06-13
Create package for using our components externally

baltpeter:
This PR lays the groundwork for a new package, `@datenanfragen/components`, which allows to use our components outside of `website`. This is meant for our upcoming [mobile](https://github.com/datenanfragen/mobile-app/) and [desktop app](https://github.com/datenanfragen/desktop-app/), but can also be used by others.

The package can be built using `yarn build-components-package`. Instead of Webpack, it uses Parcel as the bundler, which is a lot nicer (and also what I've setup already in `desktop-app`).

A few notes:

* Currently, the only exported component is the `ActWidget`. In the future, we may want to export additional components to allow people to create their own UX.
* The CSS build currently throws a lot of warnings. "Amusingly", this wall of warnings is caused by a just few lines of SCSS in the `hsb()` function, which however gets called a lot. I've tried to get rid of them but that's unfortunately not exactly easy as we're combining different build systems here. Hugo uses the old libsass, while Parcel uses the newer dart-sass. The math syntax we're using is deprecated in dart-sass. Meanwhile, the newer syntax doesn't seem to be supported in libsass. Phenomenal.

  It is [possible to get Hugo to use dart-sass](https://gohugo.io/hugo-pipes/scss-sass/#options) but that currently requires an annoying manual installation process, which I really don't want us to require. I haven't found a way to get Parcel to use libsass.
* The search bar uses dynamic imports which _should_ work but are broken for some reason. Since we're revamping that anyway in #895 and I would prefer to ultimately get rid of autocomplete.js completely, I didn't think it was worth it to put more effort into this.
* The `letter-generator` dependency currently points to a local path (which will make the build fail). @zner0L needs to merge https://github.com/datenanfragen/letter-generator/pull/14 so I can fix that.
* Finally, our icons currently don't work as `postcss-fonticons` isn't compatible with the newest postcss version. @zner0L is working on fixing that.


assets/styles/icons.scss
```diff
@@ -11,150 +11,150 @@
 
 .icon-letter:before {
     @extend %icon;
-    font-icon: 150% url('letter.svg');
+    font-icon: 150% url('../../assets/icons/letter.svg');
```
zner0L: If `postcss-fonticons` works correctly, this shouldn't be necessary.

assets/styles/main.scss
```diff
@@ -136,7 +136,7 @@ a #logo {
     overflow: hidden;
     z-index: -1;
 
-    background-image: url('img/wave-bg.svg');
+    background-image: url('../../assets/img/wave-bg.svg');
```
zner0L, line 139: Same here.

assets/styles/framework.scss
```diff
@@ -235,7 +235,7 @@ main a {
             background-size: calc(100% - 16.8px) 2px;
             &:after {
                 color: $accent-color;
-                font-icon: 80% url('external.svg');
+                font-icon: 80% url('../../assets/icons/external.svg');
```
zner0L: And here.

components-package/postcss.config.js
```diff
@@ -0,0 +1,19 @@
+const path = require('path');
+
+module.exports = {
+    plugins: [
+        // TODO: postcss-fonticons is broken.
```
zner0L: It should be fixed now.

package.json
```diff
@@ -36,23 +40,20 @@
...
-    },
-    "homepage": "https://www.datenanfragen.de",
     "dependencies": {
         "autocomplete.js": "^0.37.0",
         "brutusin-json-forms": "https://github.com/brutusin/json-forms",
         "deepmerge": "^4.2.2",
         "immer": "^9.0.12",
         "js-cookie": "^2.2.1",
-        "letter-generator": "^2.0.1",
+        "letter-generator": "/home/benni/coding/git-repos/Datenanfragen.de/letter-generator/letter-generator-v2.0.2.tgz",
```
zner0L: Remove this patch.

scripts/build-components-package.ts
```diff
@@ -0,0 +1,70 @@
...
+    fs.writeFileSync(join(package_dir, 'package.json'), JSON.stringify(pjson, null, 4));
+
+    // Copy license and attribution files over.
+    for (const filename of ['LICENSE', 'ATTRIBUTION', 'AUTHORS'])
+        fs.copySync(join(website_dir, filename), join(package_dir, filename));
+
+    fs.ensureDirSync(generated_dir);
+
+    // Build a JSON module that exports the templates.
+    const template_dir = join(website_dir, 'static', 'templates');
```
zner0L, line 47: Maybe this should fetch the templates first? I mean, this relies on a fresh run of the deploy script…

zner0L (changes requested):
This is exciting! Please look at the urls for the fonticons. I'm fine otherwise.

assets/styles/icons.scss
baltpeter: Way ahead of you. :P

assets/styles/main.scss
baltpeter: That's no a `font-icon:` rule.

zner0L (approved):

PR #904, opened 2022-06-18
Generator: Request flags for optional objection in erasure requests

baltpeter:
**This needs to be merged together with datenanfragen/data#1783!**

We now don't include an objection in erasure requests by default anymore. Instead, the user has to decide whether they also want to object and, if so, provide a justification for their objection. This is in line with the common interpretation of Art. 21(1) GDPR.

See datenanfragen/data#1780.

Once both PRs are merged, I'll update the sample letters.


src/i18n/en.json
```diff
@@ -293,6 +293,9 @@
         "data-portability": "Get data in a machine-readable format",
         "erase-all": "Erase all data",
         "erasure-data": "Data to erase",
+        "include-objection": "Additionally include an objection",
```
zner0L, line 296: I think we need to explain here against what the objection is. Like "Additionally object against interests of the controller"

zner0L (changes requested):
I we explain this more clearly, this is fine.

src/i18n/de.json
```diff
@@ -293,6 +293,9 @@
         "data-portability": "Daten in maschinenlesbarem Format erhalten",
         "erase-all": "Alle Daten löschen",
         "erasure-data": "Zu löschende Daten",
+        "include-objection": "Zusätzlich Widerspruch einlegen",
```
baltpeter, line 296:
```suggestion
        "include-objection": "Zusätzlich Widerspruch aufgrund Deiner besonderen Situation einlegen",
```

src/i18n/en.json
baltpeter:
```suggestion
        "include-objection": "Additionally include an objection based on your particular situation",
```

src/i18n/en.json
```diff
@@ -293,6 +293,9 @@
         "data-portability": "Get data in a machine-readable format",
         "erase-all": "Erase all data",
         "erasure-data": "Data to erase",
+        "include-objection": "Additionally include an objection based on your particular situation",
```
zner0L, line 296:
```suggestion
        "include-objection": "Additionally object against processing based on your particular situation",
```

src/i18n/de.json
```diff
@@ -293,6 +293,9 @@
         "data-portability": "Daten in maschinenlesbarem Format erhalten",
         "erase-all": "Alle Daten löschen",
         "erasure-data": "Zu löschende Daten",
+        "include-objection": "Zusätzlich Widerspruch aufgrund Deiner besonderen Situation einlegen",
```
zner0L, line 296:
```suggestion
        "include-objection": "Zusätzlich Widerspruch gegen die Datenverarbeitung aufgrund Deiner besonderen Situation einlegen",
```

src/i18n/en.json
```diff
@@ -293,6 +293,9 @@
...
+        "objection-reason": "Reason for your objection",
+        "objection-reason-explanation": "An objection has to be justified based on your particular situation. Explain why you specifically object to the company processing your data.",
```
zner0L, line 298: This definitely needs a more in depth explanation in a blog post. What are valid reasons for an objection? We should provide examples.

zner0L: And we should definitely tell users that it could just suffice to say that they just don't like their data being processed and that they don't need to invent a reason.

zner0L (changes requested):

src/i18n/en.json
baltpeter: I would love to do that but I don't think we can. The explanation is as in-depth as I can reliably provide. I haven't found any actual decisions on that topic and the legal commentary disagrees in important aspects.

> And we should definitely tell users that it could just suffice to say that they just don't like their data being processed and that they don't need to invent a reason.

That view is only supported by BeckOK, while others say that this doesn't suffice. *shrug*

zner0L (approved):

PR #926, opened 2022-08-29
Components/App: Support for local cache

baltpeter:

PR #927, opened 2022-08-29
Components/App: Various polishing before the live demo

baltpeter:
This is based on #926, which needs to be merged first.


baltpeter:
I cherry-picked the two commits into #928 because the conflicts were too annoying.

Issue #935, opened 2022-09-28
Decide how to integrate app into the website


zner0L:
The app, or, as I started calling it, the "tool", i.e. the new generator interface needs to go somewhere on the webpage. On `new_ux` it currently just replaces the generator of which @baltpeter always said that we want to keep it. So if we keep it like it is, which would keep a lot of links intact if we support the old linking scheme, we need to find a place for the legacy generator. Maybe just datenfragen.de/legacy?

baltpeter:
I already placed the old generator at /g, which is a bit silly but I also kind of like it.

PR #936, opened 2022-09-29
Add some simple tests for the app and proceedings page

zner0L:
The tests are still failing because the old generator is not at `/generator`, so we should decide what happens to them after #935 is resolved. Also, the app tests are not really "component" tests, I am unsure how to classify them.

Depends on #934 so that the new migration test doesn't fail, so please merge that first.

### DIFF
--- a/cypress/integration/components/app.spec.js
+++ b/cypress/integration/components/app.spec.js
@@ -1,0 +1,59 @@
+describe('Request generator tool component', () => {
+    it('sets correct request type and shows flags', () => {
+        const types = [
+            { type: 'access', buttonText: 'access', flagTexts: [] },
+            {
+                type: 'erasure',
+                buttonText: 'Delete',
+                flagTexts: ['Erase all data', 'Additionally include an objection'],
+            },
+            { type: 'rectification', buttonText: 'Correct', flagTexts: ['Correct data'] },
+            { type: 'objection', buttonText: 'direct marketing', flagTexts: [] },
+        ];
+
+        for (const type of types) {
+            cy.visit('/generator');
+
+            cy.contains(type.buttonText).click();
+            cy.generatorStore()
+                .then((state) => state.batchRequestType)
+                .should('be.equal', type.type);
+
+            cy.searchAndRequestCompanies(['Reddit', 'Datenanfragen']);
+
+            cy.generatorStore()
+                .then((state) => state.request.type)
+                .should('be.equal', type.type);
+            for (const flagText of type.flagTexts) cy.get('#app').contains(flagText).should('be.visible');
+
+            cy.contains('Send request').click();
+            cy.contains('Skip request').click({ force: true });
+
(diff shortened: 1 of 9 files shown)


baltpeter (approved):

package.json
```diff
@@ -45,6 +45,7 @@
         "build-components-package": "yarn tsm scripts/build-components-package.ts"
     },
     "dependencies": {
+        "@this-dot/cypress-indexeddb": "^1.1.3",
```
baltpeter, line 48: That's a devDependency.

cypress/integration/use-cases/migration.spec.js
```diff
@@ -34,29 +34,22 @@ const complaint = {
...
-                .then(() =>
-                    win.accessLocalForageStore('my-requests').setItem(`${reference}-custom-admonition`, admonition)
-                )
-                .then(() =>
-                    win.accessLocalForageStore('my-requests').setItem(`${reference}-custom-complaint`, complaint)
-                )
-        );
+        cy.getStore('@myRequestsStore')
+            .createItem(`${reference}-custom-admonition`, admonition)
+            .createItem(`${reference}-access`, accessRequest)
```
baltpeter, line 49: Is the order here deliberate?

cypress/integration/components/app.spec.js
```diff
@@ -0,0 +1,50 @@
...
+
+    it('sets correct request type and shows flags', () => {
+        cy.contains('Delete').click();
+        cy.generatorStore()
+            .then((state) => state.batchRequestType)
+            .should('be.equal', 'erasure');
+        cy.searchAndRequestSingleCompany('Datenanfragen');
+        cy.generatorStore()
+            .then((state) => state.request.type)
+            .should('be.equal', 'erasure');
```
baltpeter, line 14: We should make sure that this stays the case for subsequent requests.

cypress/integration/components/app.spec.js
```diff
@@ -0,0 +1,50 @@
...
+    });
+
+    it('loads pdf worker for pdf only companies', () => {
+        cy.contains('access').click();
+        cy.searchAndRequestSingleCompany('Instagram');
+        cy.window().then((win) => {
+            expect(win.pdfWorker).not.to.be.undefined;
+            // somehow it does not work with this: expect(win.pdfWorker).to.satisfy((o) => o instanceof Worker);
+            // "console.log(win.pdfWorker instanceof Worker)" logs false...
+            // time to 🦆 type...
```
baltpeter, line 37: That's not true anymore.

baltpeter (changes requested):

cypress/integration/use-cases/migration.spec.js
zner0L: Yep, because the code should work even though the OR comes after the other messages in the database.

PR #937, opened 2022-09-29
Leave old requests intact to allow us to rewind the changes

zner0L:


cypress/integration/use-cases/migration.spec.js
```diff
@@ -140,4 +140,57 @@ describe('Saved requests in the legacy database should be correctly migrated', (
                 status: 'done',
             });
     });
+
+    it('Does not migrated already migrated requests', () => {
```
baltpeter, line 144:
```suggestion
    it('Does not migrate already migrated requests', () => {
```

baltpeter (changes requested):

Issue #910, opened 2022-07-14
Cleanup from website#895


baltpeter:
In the interest of moving forward, we'll soon merge website#895 into `new_ux`.

But the following still needs to be addressed before `new_ux` can be merged into `master`:

- [ ] Decide how to integrate app into the website (do we want it on that weird `/app` page?
- [ ] If we do: [Proper meta data for app page](https://github.com/datenanfragen/website/blob/bd96720a28bdf90cea9e6da8e7541fe484be55e6/content/en/app.md).
- [ ] Support other languages than EN in app.
- [ ] `CompanySearchPage` needs to respect the privacy controls.
- [ ] Flow for requesting companies not in the database!
- [ ] Global settings page with settings for:
    - Transport medium (use suggested (default) or always override with x)
    - "Request machine-readable data in access requests" (defaults to `true`)
- [ ] [Do we need to call `renderLetter()`?](https://github.com/datenanfragen/website/blob/bd96720a28bdf90cea9e6da8e7541fe484be55e6/src/Components/App/FillRequestsPage.tsx#L84)
- [ ] Improve the "Your GDPR rights" blog post.
- [ ] Write an introductory text for the `RequestTypeChooserPage`.
- [ ] Proper "What's next" page.
- [ ] Functional "Export to calendar" button on "What's next" page.
- [ ] [Added fill fields shouldn't be marked as required.](https://github.com/datenanfragen/website/blob/bd96720a28bdf90cea9e6da8e7541fe484be55e6/src/Components/Generator/DynamicInputContainer.tsx#L43)
- [ ] Enable `noUncheckedIndexedAccess` in `tsconfig.json`.
- [ ] Fix "[object Object]" at the top of the "copy manually" text of PDF requests.
- [ ] The "add input" dropup is cut off if it has too many elements.
- [ ] Persist the state.
- [ ] Handle browser back/forward navigation.
- [ ] In the arrow thing at the top, long company names should be elided instead of spanning multiple rows.
- [ ] Collapse `MailtoDropdown` to mere `mailto:` link on mobile.
- [ ] I know I've said it before but I don't think the blue background is an intuitive way to indicate a selected state. This seems more like a hovered/focussed state. This is especially true if there is only one result (e.g. if you search for "google").

baltpeter:
We'll also need to address the user testing feedback:

- [ ] https://github.com/datenanfragen/website/pull/895#issuecomment-1177486267

zner0L:
[...]

PR #947, opened 2022-10-12
Tests for the new UX (re: #910)

baltpeter:


baltpeter:
Remaining:

* The tests sometimes fail to run in CI. Cypress complains that it can't reach the dev server.
* The command for the generator store is flaky for some reason (but the one for the proceedings store seems to work fine).
* The [object Object] test will fail until the letter-generator PR is merged and the package updated here.

baltpeter:
> The tests sometimes fail to run in CI. Cypress complains that it can't reach the dev server.

This is also happening on master and I'm really not sure what the problem is. It's _not_ a timeout problem as we've had multiple times in the past. The first n tests do pass, so Hugo must have finished building successfully. But then at some point (which seems to vary between runs), Cypress can't connect to the dev server anymore. Maybe Hugo crashes for some reason?

I've not been able to reproduce the problem locally.

baltpeter:
Couldn't even reproduce it in Circle's SSH mode. :/

baltpeter:
> The command for the generator store is flaky for some reason (but the one for the proceedings store seems to work fine).

Ok, the problem is actually quite obvious. The property is only set in a `useEffect`: https://github.com/datenanfragen/website/blob/44984086b094c8079465dc4e7179d02ddb5b97ca/src/app.tsx#L11-L16

It isn't exactly surprising that that is flaky. 

I guess, I'll have to implement a wait mechanism for that… *shrug*

baltpeter:
I've tried using a higher resource class in Circle to hopefully avoid the Hugo problem. We'll have to see whether that actually works.

That still leaves the context problem, though…

zner0L:
I was thinking that maybe using the cypress wait function like `cypress-wait-until` does instead of a timeout might help with the wait issues. But I don't want to include the packages because it is the typical of [a one file npm package](https://github.com/NoriSte/cypress-wait-until/blob/master/src/index.js). At least no dependencies. What do you think, @baltpeter?

zner0L:
I rebased and hopefully fixed the context problem. I would like you to look at my commit since this is your PR, and if it checks out you can merge, @baltpeter.

zner0L (approved):

src/Components/Generator/GeneratorTestInterface.tsx
```diff
@@ -0,0 +1,29 @@
...
+// Expose the store for the test interface
+export const GeneratorStoreTestInterface = (props: GeneratorStoreTestInterfaceProps) => {
+    const generatorStoreApi = useGeneratorStoreApi();
+
+    useEffect(() => {
+        if (process.env.NODE_ENV === 'development') {
+            (window as WindowWithGeneratorApi).generatorStoreApi = {
+                ...(window as WindowWithGeneratorApi).generatorStoreApi,
+                [props.generatorId || 'default']: generatorStoreApi,
+            };
```
baltpeter, line 24: That seems overly complicated. You're assigning anyway. Why not just…?

```ts
(window as WindowWithGeneratorApi).generatorStoreApi[props.generatorId || 'default'] = generatorStoreApi;
```

baltpeter (commented):
Did you make changes in commits other than "Use waitUntil to wait…"? If so, I didn't see them due to the rebase changes.

See below for my comments. Other than that, it's fine to merge if the tests pass now.

src/Components/Generator/GeneratorTestInterface.tsx
zner0L: That threw an error if `window.generatorStoreApi` is undefined.

src/Components/Generator/GeneratorTestInterface.tsx
```diff
@@ -0,0 +1,29 @@
...
+        }
+    });
```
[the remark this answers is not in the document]
zner0L: That was an oversight.

Issue #358, opened 2020-08-12
Do not serve all language strings in `general.js`


mal-tee:
* `general.gen.js` contains the strings for all languages, but we only need them for one.
* I would personally prefer if the strings are not in the .js at all, but in a seperate file, what do you think about that?

PR #953, opened 2022-11-04
More cleanup from #895 (re: #910)

baltpeter:


baltpeter:
For reference, I'll document my process of finding out why `commons.bundle.gen.js` has grown to 1.35 MiB and fixing that.

I started by gaining an overview of what causes the large bundle size.

1. Have webpack generate a `stats.json`: `BABEL_ENV=production yarn run --silent webpack --mode=production --config webpack.prod.js --profile --json > stats.json`
2. Use [webpack-bundle-analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer) to visualize that: `npx webpack-bundle-analyzer stats.json`

The problem is obvious:

image

Somehow, pdfmake ended up in the `commons` bundle. The question now is how.

None of the source files in this repo import anything from `pdfmake` directly, so that can't be the problem. But our `letter-generator` library does. Is that the problem?

One simple hack to find out: I just commented every import of the lib out. That of course breaks both the JS and the TS but webpack kindly doesn't mind and produces a build regardless. And indeed, we have clearly narrowed down the problem a little—pdfmake is now gone from the bundles:

image

Time to narrow it down further by commenting the imports back in bit by bit. Unsurprisingly, `pdf.worker.ts` is not the problem—that is neatly contained in its own bundle, as expected:

image

But unfortunately, re-enabling any of the other imports leads to including pdfmake, even though none of them uses any of the pdfmake functionality. Seems like our changes to the `letter-generator` build process have broken the ability to tree-shake the library. 

The problem is unfortunately more complicated than I had hoped. I'll come back to this later.

mal-tee:
Sidenote: Did we regress on #358 via the translations-dummy-bundle? (Whatever that is :D)

baltpeter:
> Sidenote: Did we regress on #358 via the translations-dummy-bundle? (Whatever that is :D)

No, that bundle isn't actually used anywhere:

https://github.com/datenanfragen/website/blob/d49cf314e08fa889082df1f9b5820bcc99f2322e/webpack.common.js#L19-L20

baltpeter:
I've moved the bundle size investigation to #957. As the problem already exists on `master`, this isn't merge-critical for `new_ux`

zner0L (approved):
Thanks!

PR #956, opened 2022-11-17
Enforce /generator as slug for all translations

zner0L:
Against our previous practice, we decided to enforce the slug of the generator page to stay `generator` in all languages, because redirects from other languages would break the URL parameters. We found this to be the least technically invasive solution. It is still possible to translate the slug in the `alias` section of the frontmatter, but note that these URLs will not support URL parameters.

Sorry to the translators who cared about the translated URLs and people who used the different links previously.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,8 @@ While you can just contribute translations through this repository, an easier wa
 
 If you're interested in chatting about our translation efforts, join our [i18n room on Matrix](https://matrix.to/#/#dade-i18n:matrix.altpeter.me).
 
+Notice to translators: When adding a new language, the `slug` of the `generator.md` page must not be translated, as our redirection code depends on it.
+
 <!-- 
 TODO:
 
--- a/README.md
+++ b/README.md
@@ -73,8 +73,6 @@ While you can just contribute translations through this repository, an easier wa
 
 If you're interested in chatting about our translation efforts, join our [i18n room on Matrix](https://matrix.to/#/#dade-i18n:matrix.altpeter.me).
 
-Notice to translators: When adding a new language, the `slug` of the `generator.md` page must not be translated, as our redirection code depends on it.
-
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -73,6 +73,8 @@ While you can just contribute translations through this repository, an easier wa
 
 If you're interested in chatting about our translation efforts, join our [i18n room on Matrix](https://matrix.to/#/#dade-i18n:matrix.altpeter.me).
 
+Notice to translators: When adding a new language, the `slug` of the `generator.md` page must not be translated, as our redirection code depends on it.
```
baltpeter, line 76: I think https://github.com/datenanfragen/website/blob/master/src/i18n/README.md would be a better place for this.

The root README only has information on how to contribute translations, but no specific guidelines.

content/es/generator.md
```diff
@@ -5,6 +5,6 @@
     "description": "Usa nuestro generador para crear solicitudes RGPD de forma fácil y gratuita y ejerce tus derechos.",
     "featured_image_url": "/img/card/generator-es.png",
     "tags": [ "datarequests", "rgpd", "reglamento general de protección de datos", "generador de solicitud", "privacidad" ],
-    "slug": "generador",
-    "aliases": ["generator"]
+    "slug": "generator",
```
baltpeter, line 8:
```suggestion
```

content/pt/generator.md
```diff
@@ -2,6 +2,6 @@
     "title": "Gerador de solicitação",
     "type": "generator",
     "static": true,
-    "slug": "gerador",
-    "aliases": ["generator"]
+    "slug": "generator",
```
baltpeter, line 5:
```suggestion
```

baltpeter (changes requested):

baltpeter (approved):

PR #970, opened 2022-12-05
Final polishing changes before we release

zner0L:


baltpeter (approved):
Thanks. Assuming you have tested that the checked highlight hack actually works (don't want to wait for the deploy preview), you can merge.

PR #988, opened 2022-12-13
SvaFinder: Be more correct with evangelical + catholic church in DE

baltpeter:
Mainly:

* Add two previously missing SVAs for the German evangelical church.
* Correct a few things about the catholic dioceses.
* Remove "Datenschutzbeauftragte des Verbandes der Diözesen Deutschlands". I misunderstood. That is an internal DPO, not an SVA.

Even with those changes, our handling of churches is still not correct. I'll open an issue for that. I've opened #989 to track that.

Thanks to Felix from artikel91.eu for making us aware of those mistakes cases!


src/Components/SvaFinder.tsx
```diff
@@ -49,15 +49,43 @@ const steps: Record<string, Steps> = {
             // this allows us to determine which ones the BfDI is not responsible for.
             'bund-kk': 'debfdi',
             kirche: {
-                ev: 'deekdbfd',
+                ev: {
+                    // Source: https://dsbkd.de/geschaeftsverteilung/
+                    evdsbkd: 'deevpfalz',
+                    // Source: https://datenschutz.ekd.de/vernetzung/andere-evangelisch/
+                    evpfalz: 'dedsbkd',
```
zner0L, line 56: Is this correct or did you switch them around?

src/i18n/en.json
```diff
@@ -834,14 +834,18 @@
         "sh": "Schleswig-Holstein",
         "thue": "Thuringia",
         "ev": "Evangelical church",
+        "ev-q": "Which church or Diakonisches Werk does your request concern?",
```
zner0L, line 837: I guess you didn't find a good translation here. What about just using "Diaconia" like [Wikipedia](https://en.wikipedia.org/wiki/Diaconia)?

zner0L (changes requested):
Nice, thank you and the helpful user who helped us keep this correct. I still have two minor questions before we merge.

src/i18n/en.json
baltpeter: I'm not well-versed with this, but I think "Diaconia" describes the theological concept rather than the institution.

baltpeter (commented):
fix mixup

zner0L (approved):

PR #993, opened 2022-12-16
Recording of the NPA 122 talk

baltpeter:


zner0L (approved):
Thanks so much, I know subtitling has been an enormous amount of work!

PR #1007, opened 2023-01-11
Re-add .nvmrc

baltpeter:
Why did we delete this again, @mal-tee?


mal-tee:
To unclutter the root folder. We also assumed that we don't need it.

baltpeter:
Well, we do indeed not need it for CI. But it is very helpful for local development.

mal-tee:
Do you mind removing the .percy file? We haven't used percy in ages, right? And with doing so we cankeep the amounts of files in root the same. :D

baltpeter:
> Do you mind removing the .percy file?

Oh, good catch. I missed that. #1013

mal-tee (approved):
We should upgrade our node version. :D

PR #1019, opened 2023-01-18
Fixes #1010: Use requestFallbackLanguage() everywhere

### LINKED ISSUE #1010: Why does resetRequestToDefault always use English if not language is passed?
While working on #1008, I stumbled across this:

https://github.com/datenanfragen/website/blob/18192edfcac895536899b6b8be0ebe6fb2d41b5b/src/store/request.ts#L270

Why is the progression `language` -> `REQUEST_FALLBACK_LANGUAGE`? Shouldn't we use the website language instead of `REQUEST_FALLBACK_LANGUAGE`? Why would we set the default request language to `en` for non-English sites?

zner0L:


baltpeter (approved):
Thanks. A test for this would be nice. :P

PR #1026, opened 2023-02-09
List PGP keys on company single pages

baltpeter:
The schema has supported this for a while (datenanfragen/data#690) but it seems like we forgot to actually display the data in the UI.


mal-tee:
Thank you!

Nitpick: What about aligning the two lines as we do with phone, fax, email, and website?

mal-tee (approved):
No we don't support keys without a fingerprint and vice versa, but I guess that is bad style anyways. :D

Thank you!

PR #1038, opened 2023-03-03
Cleanup from `webform` support

baltpeter:


mal-tee (approved):
Thank you!

src/Components/RequestList.tsx
```diff
@@ -392,7 +397,13 @@ export const ProceedingRow = (props: ProceedingRowProps) => {
                                     msg.subject && (
                                         <>
                                             <br />
-                                            <span className="icon icon-email">{msg.subject}</span>
+                                            <span
+                                                className={`icon ${iconClassForTransportMedium[msg.transport_medium]}`}
+                                                title={t('sent-via-medium', 'my-requests', {
+                                                    medium: t(msg.transport_medium, 'my-requests'),
+                                                })}>
```
zner0L, line 404: This should actually also be an `aria-label` (https://www.a11yproject.com/posts/title-attributes/).

src/Components/RequestList.tsx
```diff
@@ -358,7 +358,12 @@ export const ProceedingRow = (props: ProceedingRowProps) => {
...
                                         <br />
                                         <button
-                                            className="button button-unstyled icon icon-email"
+                                            className={`button button-unstyled icon ${
+                                                iconClassForTransportMedium[msg.transport_medium]
+                                            }`}
+                                            title={t('sent-via-medium', 'my-requests', {
+                                                medium: t(msg.transport_medium, 'my-requests'),
+                                            })}
                                             onClick={() => {
```
zner0L, line 367: Same here. This needs an `aria-label` as well.

zner0L: I guess an eslint rule would be useful for that.

zner0L (changes requested):
Looks good, just a small a11y issue.

zner0L (approved):

PR #1039, opened 2023-03-05
Suggest: Only POST if changes were made.

mal-tee:
To reduce spam. Phone and fax are compared without whitespace. This introduces `fast-deep-equal` as a new first-level dependency. It was a transitive dependency before already.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
         "autocomplete.js": "^0.37.0",
         "brutusin-json-forms": "https://github.com/brutusin/json-forms",
         "deepmerge": "^4.2.2",
+        "fast-deep-equal": "^3.1.3",
         "idb": "^7.0.2",
         "immer": "^9.0.12",
         "js-cookie": "^2.2.1",
--- a/src/i18n/de.json
+++ b/src/i18n/de.json
@@ -537,7 +537,8 @@
         "github-error": "Beim Absenden gab es ein Problem mit GitHub. Dein Vorschlag wurde nicht abgeschickt. Versuch es doch noch einmal.",
         "view-on-github": "Hier kannst Du den Fortschritt auf GitHub verfolgen",
         "ok": "Ok",
-        "similarly-named-companies": "Wir haben schon diese Firmen mit ähnlichem Namen:"
+        "similarly-named-companies": "Wir haben schon diese Firmen mit ähnlichem Namen:",
+        "no-change": "Es gibt keine Änderungen die vorgeschlagen werden können."
     },
     "schema": {
         "relevant-countries": "Relevant in Ländern",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -48,6 +48,7 @@
         "autocomplete.js": "^0.37.0",
         "brutusin-json-forms": "https://github.com/brutusin/json-forms",
         "deepmerge": "^4.2.2",
+        "fast-deep-equal": "^3.1.3",
```
baltpeter, line 51: Please add the dependency here: https://github.com/datenanfragen/website/blob/master/ATTRIBUTION

src/i18n/de.json
```diff
@@ -537,7 +537,8 @@
         "github-error": "Beim Absenden gab es ein Problem mit GitHub. Dein Vorschlag wurde nicht abgeschickt. Versuch es doch noch einmal.",
         "view-on-github": "Hier kannst Du den Fortschritt auf GitHub verfolgen",
         "ok": "Ok",
-        "similarly-named-companies": "Wir haben schon diese Firmen mit ähnlichem Namen:"
+        "similarly-named-companies": "Wir haben schon diese Firmen mit ähnlichem Namen:",
+        "no-change": "Es gibt keine Änderungen die vorgeschlagen werden können."
```
baltpeter, line 541: I think I would prefer something like:

```suggestion
        "no-change": "Du hast nichts geändert."
```

But if you want to keep this:

```suggestion
        "no-change": "Es gibt keine Änderungen, die vorgeschlagen werden können."
```

src/i18n/en.json
```diff
@@ -536,7 +536,8 @@
         "invalid-request": "Some of your entries are not in the right format or incomplete. Correct them and send your suggestion again.",
         "github-error": "There was a problem with GitHub while sending your suggestion, it was not registered. Please try again.",
         "view-on-github": "You can follow the progress on GitHub",
-        "ok": "Ok"
+        "ok": "Ok",
+        "no-change": "There are no changes to suggest."
```
baltpeter, line 540:
```suggestion
        "no-change": "You didn’t change anything."
```

src/suggest-edit.js
```diff
@@ -10,8 +10,10 @@ import { ErrorException, rethrow } from './Utility/errors';
 import { submitUrl } from './Utility/suggest';
 import { FlashMessage, flash } from './Components/FlashMessage';
 import { searchClient } from 'Utility/search';
+const equal = require('fast-deep-equal');
```
baltpeter, line 13: Please use an `import` statement.

baltpeter (changes requested):
This is very helpful, thank you!

baltpeter (approved):
Thank you!

PR #1041, opened 2023-03-05
I18N: Switch language to onChange and not onBlur

### LINKED ISSUE #977: Selecting a different language in the language modal has no effect until clicking off the element
Seen in Firefox, changing the language at the bottom of the page via the drop-down has no effect until the dropdown element is no longer in focus. This also affects the suggested language popup

To replicate:

1. Go to https://www.datenanfragen.de/
2. Select Language from dropdown: English

See that popup is not present 

3. Click elsewhere on page

See language popup modal appears

mal-tee:
Fixes #977. Previously the change would only occur on blur, making the website feel sluggish and unresponsive. This restores the pre- new_ux-behviour.

### DIFF
--- a/src/Components/I18nWidget.tsx
+++ b/src/Components/I18nWidget.tsx
@@ -27,7 +27,7 @@ export const I18nWidget = (props: I18nWidgetProps) => {
         ),
         negativeText: <Text id="stay" />,
         onNegativeFeedback: () => {
-            setNewLanguage(savedLocale);
+            setNewLanguage(savedLocale); // TODO: Is this call needed?
             dismissModal();
         },
         hasDismissButton: false,
@@ -71,7 +71,7 @@ export const I18nWidget = (props: I18nWidgetProps) => {
                         <Text id="language" />
                     </h2>
                     <div className="select-container">
-                        <select value={savedLocale} onBlur={changeLanguage}>
+                        <select value={savedLocale} onChange={changeLanguage}>
                             <option value={savedLocale}>
                                 <Text id={`language-desc-${savedLocale}`} />
                             </option>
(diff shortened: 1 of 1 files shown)


baltpeter:
@zner0L removed the `onChange` handler in <https://github.com/datenanfragen/website/commit/faf6047ad0a362abda97484c5d412fe870a50d68#diff-71f2af39dee31bc7a585416ec40fb211d19629f25bae906a9c05a43360cee347R74>. Unfortunately, I don't remember why, so I'll assign him.

src/Components/I18nWidget.tsx
```diff
@@ -27,7 +27,7 @@ export const I18nWidget = (props: I18nWidgetProps) => {
         ),
         negativeText: <Text id="stay" />,
         onNegativeFeedback: () => {
-            setNewLanguage(savedLocale);
+            setNewLanguage(savedLocale); // TODO: Is this call needed?
```
zner0L, line 30: Yes, it is, but I was confused, too, so maybe let's make this a little more verbose:
```suggestion
			// Reset the select to show the current language instead of the just selected one.
            setNewLanguage(savedLocale); 
```

zner0L (dismissed):
Before I changed that to `onBlur`, we had a `onBlur` and a `onChange` handler on this because of [a eslint a11y rule](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-onchange.md). But that caused some strange behavior when I tested it with the keyboard, which I think we never really did before, so I changed it. The rule had been deprecated because now, browsers behave to spec and use `onChange` correctly, so we can go back to writing spec compliant code as well :) Thanks for correcting this, @mal-tee.

zner0L (approved):

baltpeter:
@zner0L You broke the deploys on master…

PR #1048, opened 2023-03-20
Fix whitespace issue

mal-tee:


mal-tee:
Sometimes I miss reviewbot. :D

baltpeter:
Wouldn't be too hard to set that up here as well. :D

mal-tee:
\o/ go go go go go go :D

PR #1075, opened 2023-06-10
Update browserlists

mal-tee:


baltpeter (approved):
Thank you very much for this very important change that will propel our project into the next century.

PR #1100, opened 2023-09-18
Annotated presentation for Topio 2023 talk

baltpeter:
Inspired by https://simonwillison.net/2023/Aug/6/annotated-presentations/

This isn't quite done yet, @zner0L's new slides don't have text yet.


mal-tee (commented):
This is great! But I am a bit torn on the design, why don't we wrap this in a slide show?

baltpeter:
I adapted Simon's design for the most part. I do like having everything on one page for discoverability (scrolling is about as obvious as it gets). We do have a JS-free implementation of a carousel for the Honey act page but that is annoying to produce and… not great. Also, I'm not sure how it holds up a11y-wise.

Why would you prefer a carousel, what benefits do you see?

mal-tee:
> Why would you prefer a carousel, what benefits do you see?

Do separate the slides from the background of the page. I guess a border or some other CSS styling could fix this as well. I'd also prefer a bolder line between slides. Maybe even introduce a new shade for the slide's caption?

And I am once again proposing that we should bump the image quality. Imo these slides are compressed way too much.

baltpeter:
> Do separate the slides from the background of the page. I guess a border or some other CSS styling could fix this as well.

But why? We don't separate "regular" images (e.g. in blog posts) from the background, either? :D 

> I'd also prefer a bolder line between slides.

Imo that looks silly but I don't care that much. What does @zner0L think?

PR #1103, opened 2023-09-19
Suggest: Hide required elements and comments

baltpeter:
We never know whether people are just guessing for the required elements and usually end up removing them anyway for that reason.

And as for the comments: Records should only have comments in very rare cases. This field just encouraged people to leave the kinds of comments that should be left on the company pages instead.


baltpeter:
Exactly, I'm fine with having a free-form comment field that just posts comments in the GitHub PR. But having those go to the record comments is just wrong.

zner0L:
Would you mind opening an issue for that then, so that we can implement this later?

mal-tee:
If we postpone the new comment feature, we will never implement it. A PR that disables the status quo comments should introduce a new comment feature as well imo.

christian-weiss:
Do i get it right, the following change is requested:

- [ ] Update "PR creator" to include the comments in PR text
- [ ] drop "comments" from "file creator" (for PR) (there is no need to persist the comments in company records)
- [ ] drop "comments" from schema.json (there is no need to persist the comments  in company records)
- [ ] update "comments" field from all records in "data" repo to make them fit to the new schema
- [ ] drop code that read/write/access the "comments" field

right?

mal-tee (approved):
After two years of intense considerations, I have changed my mind. ;)

PR #1116, opened 2023-11-27
Show helpful error for comments that are too long

baltpeter:
This is for https://github.com/datenanfragen/backend/pull/79. This PR should be deployed first.

### DIFF
--- a/src/Components/CommentsWidget.tsx
+++ b/src/Components/CommentsWidget.tsx
@@ -4,7 +4,7 @@ import { IntlProvider, Text, MarkupText } from 'preact-i18n';
 import { FlashMessage, flash } from '../Components/FlashMessage';
 import { StarWidget } from './StarWidget';
 import t from '../Utility/i18n';
-import { rethrow, WarningException } from '../Utility/errors';
+import { ErrorException, rethrow, WarningException } from '../Utility/errors';
 import { useAppStore } from '../store/app';
 
 const api_url = 'https://backend.datenanfragen.de/comments';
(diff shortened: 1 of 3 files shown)


src/Components/CommentsWidget.tsx
```diff
@@ -226,8 +226,12 @@ export function CommentForm(props: CommentFormProps) {
                 ...(props.allowRating && rating ? { additional: { rating } } : {}),
             }),
         })
-            .then((res) => {
-                if (!res.ok) throw new Error('Unexpected response from comments server.');
+            .then(async (res) => {
+                const payload = await res.json();
+
+                if (payload.message?.includes('length must be less than')) {
```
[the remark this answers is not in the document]
baltpeter: No.

mal-tee, line 232: So we should do this instead?

```suggestion
                if (payload?.message?.includes('length must be less than')) {
```

baltpeter: But `payload.message?` is already guarding against a payload potentially not having a message?

Your suggested change would guard against `payload` itself being `null` or `undefined`. If the backend is behaving correctly, we do always send a JSON payload with a `message` (https://github.com/datenanfragen/backend/blob/b_length-limits/src/handlers/comments/putComment.js#L11, https://github.com/datenanfragen/backend/blob/b_length-limits/src/handlers/comments/putComment.js#L32, https://github.com/datenanfragen/backend/blob/b_length-limits/src/handlers/comments/putComment.js#L36).  
The only way for `payload.message` to throw is if the response body was `null`. If that happens, the backend is clearly malfunctioning and we would drop into the `catch` anyway, which shows an appropriate error message.

mal-tee (approved):
Thank you!

Issue #1120, opened 2023-12-17
mollie integration exception


mal-tee:
We got a user report about a broken donation with mollie:

> Uncaught Critical: Malformed response body when trying to get payment provider auth URL.

Since its donation season/end of the year, we should look into this? :D

mal-tee:
I can reproduce this by selecting mollie and clicking continue.

baltpeter:
We have completely disabled Mollie payments in https://github.com/datenanfragen/website/pull/1177, so this can be closed.

PR #1121, opened 2023-12-18
Disable mollie until we fix the backend. ref. #1120

mal-tee:


zner0L (approved):

zner0L (changes requested):
Prettier is unhappy, can you maybe fix the problem? Otherwise I agree, thank you :)

PR #1132, opened 2024-01-17
CI: No useless chmod, switch ENV from array to object

christian-weiss:
see commit messages


.circleci/config.yml
```diff
@@ -46,7 +45,6 @@ jobs:
           command: |
             export CONTEXT="production"
             yarn install
-            chmod +x deploy.sh
             ./deploy.sh
```
davenewham, line 48: I think this command will no longer work if the file is not marked as executable? It wil need `sh deploy.sh` if it's necessary to not mark this as executable

davenewham: Ignore my comment, the deploy.sh file is already committed with the executable flag :)

baltpeter (approved):
Thanks.

PR #1145, opened 2024-03-22
Comments: Expand regex for detecting wrongly addressed comments

baltpeter:
Based on a few comments we have received since the original PR was merged.

Two more changes in this commit:
* Also calculate a "sus score" for the "author" field, we get quite a few comments with message text in that field.
* Accordingly, do the "no email in author" check first.


mal-tee (approved):
Great!

PR #1154, opened 2024-06-18
Error-Handler: Ignore errors if the browser is outdated. Fixes #644

### LINKED ISSUE #644: Show warning if users are using a deprecated browser?
Sometimes we get error reports of users using weird or very old browsers. Should we show a warning for unsupported browsers?

On the one hand this might get users to switch to a supported browser, on the other they might have no choice and could be annoyed by such a warning. But if it's easily dismissable that shouldn't be much of a problem.

Integration into `browserslist` or something similar would be great to determine when to show the warning. Otherwise we could make a list of browsers we know are not supported.

If we decide on doing this, this could make a good hacktoberfest issue. 😄

mal-tee:


baltpeter:
Thanks! But my concern is that we should only prevent users with unsupported browsers from reporting errors, not from seeing them. That would only lead to the site being inexplicably broken for them.

My suggestion would be to still show the modal but not the sections for reporting the error but instead display a warning that this is likely caused by an outdated browser.

mal-tee:
> Thanks! But my concern is that we should only prevent users with unsupported browsers from reporting errors, not from seeing them. That would only lead to the site being inexplicably broken for them.

But we are doing the same with extension users. Should we change the behaviour there as well? But I don't think we really want to support debugging weird extensions..

baltpeter:
The extension errors usually are meaningless for us (they only occur in the extensions' content scripts) and don't affect our site which continues to work fine.

That's not the case here, though. If the user's browser doesn't support the functions you check for, features on our site **will** break. If we don't show any notice to the user, the site will just be inexplicably broken for them. We should tell them that the error was caused by their outdated browser and that they need to switch to a supported browser to use our site.

src/error-handler.js
```diff
@@ -82,6 +98,21 @@ const isNetworkError = (message) => {
     if (message === 'NetworkError when attempting to fetch resource.') return true;
     return false;
 };
+
+/**
+ * Heuristic to determine if the browser is too old. Uses a somewhat recent language feature to check if it is supported.
```
mal-tee, line 103:
```suggestion
 * Heuristic to determine if the browser is too old. Uses somewhat recent language features to check if the browser is deprecated.
```

PR #1170, opened 2024-11-15
Set COOP header

mal-tee:
Inspired by #322 and this [talk](https://media.ccc.de/v/god2024-56286-modern-solutions-against-c): let's bring our security headers up to speed. This one mitigates cross-site leaks.


baltpeter:
Thank you! We do have one instance where we use `window.open()` to create a window with `url: ''` and write content into that:

https://github.com/datenanfragen/website/blob/05b4f828f7450bcf25e2a70b4514deb60c5484aa/src/Components/DonationWidget.tsx#L360-L373

From watching the talk you linked and glancing over https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Opener-Policy, I'm not actually sure whether this would be affected. Can you confirm/research whether this would be a problem?

mal-tee:
It works in the deploy preview. My (naive) understanding is that although `""` is cross-origin from us, `""` doesn't set the COOP.

Dade setting the header doesn't affect this case, as it is not dade which is opened in the popup.

baltpeter:
Oh right, I totally forgot that `_headers` also works on Netlify (which is where I stole that from :D). Coulda tested that myself then. Thank you! :D

baltpeter (approved):

PR #1179, opened 2025-05-11
Direct users trying to object to Meta "AI" (incl. temp hack for FR)

baltpeter:
We have received a lot of French comments from people trying to object to Facebook using their data for "AI" training. Since this is a current event, we'll try to direct them to the right place until July 26, 2025.


mal-tee (approved):
LGTM. We could improve this by making the box more prominent and establishing a better process, but this is not a blocker for this "event".

PR #1189, opened 2025-08-21
Optimize deploy script by only renaming once

baltpeter:
With more languages and companies being added, our deploy script has been getting slower and slower for a while. That makes sense since we have to copy and rename each record for each language. With currently more than 3,000 companies and 8 languages, that is quite a lot of work.

Today, I happened to notice that the copying of the files is super fast while the renaming takes a long time. Taking a closer look at the script, that shouldn't be too surprising: We're spawning a separate shell for each rename operation, ~24,000 shells in total.

There's a trivial optimization to be had here: Only do the renaming once and _then_ copy the already renamed files. While still not great (since we still spawn ~3,000 shells), that is a huge improvement.

On my machine, the previous version of the deploy script took 56s excluding the "Running Webpack and Hugo…" steps. With this change, it only takes 12s.

And since we're literally just rearranging code, this should also be fine in CI, unlike our earlier tries with GNU parallel.

---

From looking at the code, I'm _reasonably_ sure that this change shouldn't break anything (like the JSONs in `static/db`), but it would be good to have this reviewed carefully.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -39,6 +39,13 @@ echo "Copying files…"
 cp data_tmp/companies/* static/db
 cp data_tmp/supervisory-authorities/* static/db/sva
 
+# Unfortunately, Hugo only accepts .md files as posts, so we have to rename our JSONs, see https://stackoverflow.com/a/27285610
+# Since we're spawing a shell for each file, this is fairly expensive and considering we have thousands of company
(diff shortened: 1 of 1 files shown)


baltpeter (commented):
I'm not going to claim to have understood the xargs and sed spells, but they look reasonable and do work. And I can't deny that this is now a lot faster, so *yay*. :D Approved.

PR #1190, opened 2025-08-25
Handle obsolete records

baltpeter:
This PR is the counter part to https://github.com/datenanfragen/data/pull/3204.

The build currently fails because the changes expect the data PR to be merged. I have tested it both locally and on Netlify (temporarily change the deploy script to check out the `b_obsolete-records` branch, i.e. `git clone --depth 1 -b b_obsolete-records https://github.com/datenanfragen/data data_tmp`).


baltpeter:
Well, now I'm gonna accuse you of cherrypicking. :P As per the [guidelines](https://github.com/datenanfragen/data/#cases-where-there-is-still-a-controller) the name for the company/service that users would use to search for it should be added to the `runs` array. And I did do that in most cases where there is a usable source (finding such sources is a problem that is harder than one might think…).

`aachen-muenchener` is an (annoying) exception for two reasons:

1. When I added the obsolete record, `generali` did in fact still list `"AachenMünchener Versicherung AG (ehem.)"`  in `runs` (https://github.com/datenanfragen/data/commit/cc8ab865ce8305c5010ab74cf9f90d1eaeccd6e4#diff-b8d2b30c89e55ff6b546b395d963cdbb84f0f693bf515dda63bbc7f048a2f522L15). But while that PR was waiting to be merged, in an unrelated commit while updating outdated records, I noticed that Generali is one of those annoying cases where we had previously wrongly listed a bunch of companies under `runs` that actually need to be treated as separate controllers and thus need separate records. In that commit (https://github.com/datenanfragen/data/commit/cc8ab865ce8305c5010ab74cf9f90d1eaeccd6e4), I thus split the record. In the process, I removed the old AachenMünchener references.
2. The reason for the removal of AachenMünchener is that I'm really not sure which of those companies is responsible for the old AachenMünchener data. If you manage to find that out and also find a suitable source (that includes the string you want to insert into `runs`, for future `keepalive` stuff), please absolutely do add it back.[^aside]

[^aside]: Side note: I think that the bar for redirecting an obsolete record to a new one is lower than for including an item in `runs` (which will be included in the requests to the company, claiming that they are responsible for it). Thus, I am fine with redirecting `aachen-muenchener` to `generali`, even though I am not sure whether that is the correct record. But if people disagree, I'm also fine with removing the `redirect-to` here.

That being said, I do agree that we should inform users why they were redirected.

PR #1205, opened 2025-10-18
Node 24 Compatibility

davenewham:
**Summary of changes**

- Replace tsm for tsx
- Replace node-sass for sass
- Update glob to glob 13 and update use globSync instead of glob.sync
- Update CI node versions to use v24.13.0

These changes were all trivial, now that webpack is out of the equation.

**Testing environment:**
Debian Node v24.13.0, Hugo 140
Two commits were cherry picked into the branch to test ./deploy.sh and build - otherwise the build would locally fail

The CI build failures appear to be pre-existing as seen on the original PR #1211 so I will leave those for @baltpeter to investigate

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -81,7 +81,3 @@ else
     cp _headers public/_headers
     cp _redirects public/_redirects
 fi
-
-# Finds all generated css files, matches and removes the second last non-dot characters (the md5 hash) and renames the files to the new filename without hash
-# This is really not a good fix and I beg hugo to change this!
-find "public" -regex '.*/styles/.*\.css' -print | sed -e "p" -e "s/\(.*\.min\)\.[^\.]*\(\.[^\.]*\)$/\1\2/" | xargs -P $process_number -n 2 mv


baltpeter:
Awesome, thank you, @davenewham!

> Replace node-sass for sass

I was worried that this might break things because it has for me it the past. After some investigation, I found why it doesn't. We're not actually using that package anywhere. :D 

It was added when we were still building the CSS with Webpack (https://github.com/datenanfragen/website/commit/54cf1f26944b82896c13f5909274b98fe6f26789). We've long since migrated to building the CSS with Hugo (https://github.com/datenanfragen/website/pull/375). Hugo ships with an embedded LibSass (<https://gohugo.io/functions/css/sass/>).[^1]

[^1]: The embedded LibSass has been deprecated as of https://github.com/gohugoio/hugo/releases/tag/v0.153.0 (in a very easy to miss note), but will be kept around for a while (https://github.com/gohugoio/hugo/issues/14261). In any case, Hugo doesn't appear to use `sass` installed via NPM as per my testing, it needs to be installed globally (https://github.com/gohugoio/hugo/issues/14261).

The build for the components package uses @parcel/transformer-sass, which depends on `sass` itself (https://github.com/parcel-bundler/parcel/blob/59484858a1a0bcbb71f74088956bb437a2db6505/packages/transformers/sass/package.json#L25). And as expected, that build indeed throws tons of deprecation warnings. :D 

Anyway, we don't need to depend on `sass`. I have thus removed the dependency.

> The CI build failures appear to be pre-existing as seen on the original PR https://github.com/datenanfragen/website/issues/1211 so I will leave those for @baltpeter to investigate

Oops, it's embarrassing that we didn't catch that in #1212. But it's an easy fix. We need to set the `sourceMap` option of Hugo's `js.Build` to `"none"` instead of `false` (https://gohugo.io/functions/js/build/#sourcemap). Fixed.

> Two commits were cherry picked into the branch to test ./deploy.sh and build - otherwise the build would locally fail

I'll rebase https://github.com/datenanfragen/website/tree/webpack-node-staging and this PR on the current master.

baltpeter:
Oops, and another build failure. We really didn't review #1212 carefully, did we? :| Since we don't fingerprint our CSS assets anymore (https://github.com/datenanfragen/website/pull/1212/changes/09dff91fd2ee7b7ac0a48cf6560a03440f329859), the moves in `deploy.sh` fail. I'll fix that as well.

Issue #1167, opened 2024-11-06
Disable password manager autocomplete on suggest edits


zner0L:
To prevent users from accidentally committing private data to the database, we should add custom attributes to the inputs to indicate that password managers and browsers should not fill the user’s information in. This Stackoverflow question describes how: https://stackoverflow.com/questions/41945535/html-disable-password-manager

PR #1208, opened 2025-10-21
feat(suggest): disable autocomplete

Jolg42:
Closes https://github.com/datenanfragen/website/issues/1167

Set attributes to avoid autocompletion on the form This is to avoid private data to be submitted (e.g. via a Password Manager)

### DIFF
--- a/src/suggest-edit.js
+++ b/src/suggest-edit.js
@@ -155,11 +155,16 @@ function renderForm(schema, company_data = undefined) {
 
     // Set attributes to avoid autocompletion on the form
     // This is to avoid private data to be submitted (e.g. via a Password Manager)
-    document.querySelectorAll('form.brutusin-form').forEach((el) => {
+    document.querySelectorAll('form.brutusin-form input, form.brutusin-form textarea').forEach((el) => {
         el.setAttribute('autocomplete', 'off');
-        el.setAttribute('data-1p-ignore', 'true');
+        // https://developer.1password.com/docs/web/compatible-website-design/#ignore-offers-to-save-or-fill-specific-fields
+        el.setAttribute('data-1p-ignore', '');
+        // https://support.lastpass.com/s/document-item?language=en_US&bundleId=lastpass&topicId=LastPass%2Fc_lp_prevent_fields_from_being_filled_automatically.html&_LANG=enus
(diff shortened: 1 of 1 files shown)


src/suggest-edit.js
```diff
@@ -145,11 +145,23 @@ function renderForm(schema, company_data = undefined) {
...
+    // Create and render form
     bf = BrutusinForms.create(schema);
     bf.render(
         document.getElementById('suggest-form'),
         company_data || (PARAMETERS['name'] ? { name: PARAMETERS['name'] } : {})
     );
+
+    // Set attributes to avoid autocompletion on the form
+    // This is to avoid private data to be submitted (e.g. via a Password Manager)
+    document.querySelectorAll('.brutusin-form form').forEach((el) => {
```
mal-tee: This selector returns nothing, if I am not mistaken. Did you mean `input`? As far as I understand the attributes should be set in the `input` field anyways.

Did you test this with a password manager? :) 

```suggestion
    document.querySelectorAll('.brutusin-form input').forEach((el) => {
```

Jolg42: From the MDN documentation, they can also be set on the form and not repeated everywhere.
Maybe it's better to try this version, if it works for all as expected then it's better I think 🤔 

https://developer.mozilla.org/en-US/docs/Web/Security/Practical_implementation_guides/Turning_off_form_autocompletion
> You can do this either for an entire form, or for specific input elements in a form.

baltpeter:
I'm using 1Password and just gave this a quick test. Since the attributes currently aren't actually being set as @mal-tee said, I manually added them using the Firefox dev tools. The autocomplete still appeared both when setting `data-1p-ignore` on the `form` as well as an `input`. So, this could either be a bug in 1Password _or_ the attributes have to be set already when the form is loaded and can't be added later (which would make sense I guess if 1PW has already injected its scripts into the form).

src/suggest-edit.js
Jolg42: Good find! I thought I saved and added that change to my commit, but I did not apparently 😄

baltpeter:
## Firefox native autofill

On master:

image

With the attributes set on the individual elements, it still wants to autofill…

image

Setting `autocomplete="off"` on the form itself doesn't work, either:

image

PR #1209, opened 2025-10-21
Fixes dependency version for brutusin-json-forms to specific commit hash

Jolg42:
While working on https://github.com/datenanfragen/website/pull/1208 I saw that the brutusin-json-forms is not pinned and outdated since 7 years.

This pull request makes a minor update to the `brutusin-json-forms` dependency in `package.json`, pinning it to a specific commit hash for improved stability and reproducibility.

It's slightly better than the default branch on a GitHub repo :)

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3112,7 +3112,7 @@ browserslist@^4.20.2:
     node-releases "^2.0.3"
     picocolors "^1.0.0"
 
-"brutusin-json-forms@https://github.com/brutusin/json-forms":
+"brutusin-json-forms@https://github.com/brutusin/json-forms#44f27b29ef657f545b8a3d162c2b9c90ef137dbc":
   version "0.0.0"
   resolved "https://github.com/brutusin/json-forms#44f27b29ef657f545b8a3d162c2b9c90ef137dbc"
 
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -3112,7 +3112,7 @@ browserslist@^4.20.2:
     node-releases "^2.0.3"
     picocolors "^1.0.0"
 
-"brutusin-json-forms@https://github.com/brutusin/json-forms":
+"brutusin-json-forms@https://github.com/brutusin/json-forms#44f27b29ef657f545b8a3d162c2b9c90ef137dbc":
   version "0.0.0"
   resolved "https://github.com/brutusin/json-forms#44f27b29ef657f545b8a3d162c2b9c90ef137dbc"
```
mal-tee: Thanks for the PR! But doesn't the `resolved` entry already pin the version? :)

Jolg42: Yes indeed, but it makes the dependency requirement clearer directly in `package.json`
To take an example with any other package, no package is set with `"mypackage": "latest"` in the `package.json` (though they are pinned in the lockfile).
Here the same logic applies.

Jolg42:
@baltpeter I had no idea, nice to see these issues/discussions!

Issue #554, opened 2021-02-22
Unexpectedly long deploy times


baltpeter:
I have started looking into the unusually long deploy times that we are seeing. We had previously seen something similar while still using Netlify but were able to fix that (#320). When building dattel, I made sure to also only upload the changed files, which is why deploy times only increased by a bit when we switched.

However, we are now back to around 25 mins per deploy (and have been for a while), which suggests that we have once again introduced something that causes a lot of files to change with each deploy. In fact, I was able to find what causes this problem but I am not sure how we should go about fixing it.  
Unfortunately, the problem is actually even more severe than annoyingly long deploy times. Currently, we are invalidating the cache for _all_ HTML files with each deploy.

The problem lies in the fact that our CSS URLs look like this since we started building the CSS with Hugo (#375):  
`https://www.datenanfragen.de/styles/main.min.0af338caff4de77e5b710e1847fbeceadbf12bb3bcf4cd41bac2069a09729282.css`

I have actually learned that this is generally a bad idea [from Netlify](https://community.netlify.com/t/support-guide-how-does-netlify-s-cdn-handle-caching-files/103) when researching how to implement caching for dattel. With this approach, every change to the CSS not only causes us to need to re-upload the CSS file during the deploy (and all visitors to re-download it) but this actually applies to _all_ HTML files as they all obviously reference this URL (which changes).

Thus, we know that there are actually two problems that we need to tackle:

1. Why is the generated CSS file different with each deploy?
   
   The reason for this problem is actually quite easy to find out. All we need to do is generate the CSS twice and then run a diff. From that we can see that the problem is confined to a single section:

   ![image](https://user-images.githubusercontent.com/4048809/108752415-5684dc00-753b-11eb-9c57-9bdce076cb94.png)

   For some reason, the iconfont generation leads to a different result each time. I have not investigated this further yet. The iconfonts are generated by [zner0L/postcss-fonticons](https://github.com/zner0L/postcss-fonticons) (forked from [jantimon/iconfont-webpack-plugin](https://github.com/jantimon/iconfont-webpack-plugin)). @zner0L: Any idea what we can do here?

2. How can we convince Hugo to generate the permalink for the CSS without the cache-busting filename?
   
   While I would definitely like to solve 1., a hotfix would be to fall back to the old approach of using a static CSS filename (simply `main.min.css`). Actually, I think we should do that either way due to the reasons described by Netlify. However, I have so far not found a way to do that.

[...]

PR #1212, opened 2025-10-25
Fixes #1211: Build JS with Hugo instead of Webpack

### LINKED ISSUE #1211: Build JS with Hugo instead of Webpack
We are currently building `website` with Webpack 4, which is very outdated and requires Node 14, which is ancient. Migration to Webpack 5 is somewhere between non-trivial and hard. This problem has thwarted efforts to upgrade to newer versions of Node for quite a few years now (cf. #986, #1076, #1147, #1205).

Meanwhile, we have had plans to switch away from Webpack for even longer. In #701, @zner0L attempted to switch to Hugo's [`js.Build`](https://gohugo.io/functions/js/build/) based on esbuild. That would mainly have two advantages: The configuration is much easier and more ergonomic than Webpack. And builds are _much_ faster.

The #701 PR is now very outdated and has even been created before our major refactoring and TS conversion in 2022. As such, I don't think it makes sense to build on that PR. It will be much easier to start anew and copy still relevant bits from the PR.

So, I guess that's what I'm going to attempt now. :D

baltpeter:
This should be squashed when merging.

### DIFF
--- a/scripts/build-i18n.ts
+++ b/scripts/build-i18n.ts
@@ -5,7 +5,7 @@
  * It supports building for production (without any flags) or watch mode for development (with `--watch`). It is run by
  * `yarn dev` and `yarn build`.
  */
-import { readFileSync, writeFileSync, mkdirSync } from 'fs';
+import { readFileSync, writeFileSync, mkdirSync, unlinkSync } from 'fs';
 import { join, basename } from 'path';
 import glob from 'glob';
 import { getDirname } from 'cross-dirname';
@@ -25,11 +25,13 @@ const inputDir = join(dirname, '..', 'src', 'i18n');
 const hugoOutputDir = join(dirname, '..', 'i18n');
 const jsOutputDir = join(dirname, '..', 'static', 'js');
 
-const allTranslations = Object.fromEntries(
-    glob
-        .sync('*.json', { cwd: inputDir, absolute: true })
-        .map((p) => [basename(p, '.json'), JSON.parse(readFileSync(p, 'utf8')) as TranslationFile] as const)
-);
+// We need to recompute this every time because translations may have changed in watch mode.
+const allTranslations = () =>
+    Object.fromEntries(
+        glob
+            .sync('*.json', { cwd: inputDir, absolute: true })
+            .map((p) => [basename(p, '.json'), JSON.parse(readFileSync(p, 'utf8')) as TranslationFile] as const)
+    );
 
 // eslint-disable-next-line no-console
 const log = (...messages: unknown[]) => !quiet && console.log('[i18n]', ...messages, `(${new Date().toISOString()})`);
@@ -38,7 +40,7 @@ const log = (...messages: unknown[]) => !quiet && console.log('[i18n]', ...messa
  * Emit the Hugo and JS translations for the given input translations.
  */
 const emitTranslations = (translations: Record<string, TranslationFile>) => {
-    const en = allTranslations['en'];
+    const en = allTranslations()['en'];
 
     // Ensure output directories exist.
     mkdirSync(hugoOutputDir, { recursive: true });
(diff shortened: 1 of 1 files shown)


baltpeter:
Alright. As far as I can tell, this is now ready for review. A couple of things to note:

* This should be reviewed very carefully. There is obviously quite a lot of potential for breakage. In addition to that, we also need to be on the look-out for non-obvious gotchas like https://github.com/datenanfragen/website/blob/b_hugo-esbuild/layouts/partials/scripts.html#L10-L11. I would appreciate a review from both of you, @zner0L and @mal-tee.
* As we discussed in https://github.com/datenanfragen/website/issues/1211#issuecomment-3460951828, I have been ignoring CI failures because of interdependencies between this PR and #1205, where neither can really move forward without the other. The random build failures I've been seeing (both locally and in CI) seem like a Hugo bug (locally, I've been using the latest Hugo version today and not seen it happen anymore so far), but because of our ancient Node version and which Docker images are available for it, we cannot upgrade Hugo yet. 

  As such, I propose that we merge this PR into the `webpack-node-staging` branch for the time being. Then @davenewham can rebase #1205 onto that, which should hopefully simplify that PR greatly. We'd then also merge #1205 into `webpack-node-staging`. And finally, I would work on `webpack-node-staging` and fix the final remaining issues and create a PR to merge that into `master`.
* Given the previous point, I also haven't fixed all Cypress errors yet. There are a handful when I run the tests locally, but at a glance these all seem like Cypress flake to me. Thus, I don't think it makes sense to work on them before we've updated everything.

mal-tee:
We can also bump cypress after the node bump, which will introduce some QoL features iirc :D

layouts/partials/scripts.html
```diff
@@ -1,88 +1,202 @@
...
+        {{ .SetOptions (dict "resource" (resources.Get "js/app.tsx")) }}
+    {{ end }}
 {{ end }}
 
-{{ if eq .Type "my-requests" }}
-<script src="{{ "js/my-requests.bundle.gen.js" | absURL }}"></script>
-<script>window.renderMyRequestsWidget();</script>
+{{ with $batch.Group "advanced-generator" }}
+    {{ with .Script "generator" }}
+        {{ .SetOptions (dict "resource" (resources.Get "js/generator.tsx")) }}
```
zner0L: I guess more elegantly even, we would include [a runner](https://gohugo.io/functions/js/batch/#runner) to render the components instead of doing it in the files that define them themselves.

layouts/partials/scripts.html
```diff
@@ -1,88 +1,202 @@
...
+{{ if in $groups "donate" }}
+    <script type="module">window.renderDonationWidget();</script>
+
+    {{/* TODO: Do this in the js build via hugo pipes */}}
+    {{ $loader := resources.Get "styles/loader.scss" | toCSS (dict "enableSourceMap" .Site.Params.devMode) | postCSS (dict "config" "postcss.config.js" "noMap" true) | minify | fingerprint }}
+    {{ $link := $loader.RelPermalink | absURL }}
+    {{ if (or (eq hugo.Environment "production") (eq hugo.Environment "staging")) }}
+    {{ $link = $link | replaceRE "\\.[^\\.]*(\\.[^\\.]*)$" "$1" }}
+    {{ end }}
+    <link rel="stylesheet" href="{{ $link }}" integrity="{{ $loader.Data.Integrity }}" crossorigin="anonymous">
```
zner0L: I know I wrote this a s a response to https://github.com/datenanfragen/website/issues/554, but this is stupid. If we just don’t do the fingerprinting (which we don’t need anyways) then the file names are cache friendly:
```suggestion
    {{ $loader := resources.Get "styles/loader.scss" | toCSS (dict "enableSourceMap" .Site.Params.devMode) | postCSS (dict "config" "postcss.config.js" "noMap" true) | minify  }}
    <link rel="stylesheet" href="{{ $loader.Permalink }}" integrity="{{ $loader.Data.Integrity }}" crossorigin="anonymous">
```

We should also apply this to the other changes I made in https://github.com/datenanfragen/website/pull/555/changes/8607dfe4fa2096ba20fd1330a6732c30f9613e34.

scripts/build-i18n.ts
```diff
@@ -0,0 +1,145 @@
...
+import { watch as chokidar } from 'chokidar';
+
+type TranslationScope = string;
+type TranslationKey = string;
+type TranslationFile = Record<TranslationScope, Record<TranslationKey, string>>;
+
+const watch = process.argv.includes('--watch');
+const quiet = process.argv.includes('--quiet');
+
+const dirname = getDirname();
```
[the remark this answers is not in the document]
baltpeter: Because it doesn't work in ESM.

PR #1245, opened 2025-12-28
App: Add icons to request type chooser

mal-tee:


mal-tee:
CI failure is unrelated.

PR #1268, opened 2026-04-13
WIP: Port /suggest to formj

mal-tee:


src/Utility/suggest.ts
```diff
@@ -1,3 +1,17 @@
+import Ajv from 'ajv';
```
baltpeter, line 1: Actually, now that I've thought about this a little, I don't think we should pull in ajv. For one, it's quite heavy (https://bundlejs.com/?q=ajv%408.18.0%2Cajv-formats%403.0.1%2Cajv-formats-draft2019%401.6.1). But more importantly, we don't _want_ the suggest form to do validation. The current one doesn't do that either because we've decided that the suggest form is explicitly for users who don't want to deal with the intricacies of our database.

I think we should instead add an option to formj to disable validation.

baltpeter: To get the bundle size advantage, we have to make sure that formj's built-in ajv isn't bundled.

mal-tee:
@baltpeter How does one get the Form generated by FormJ to submit? Creating an ApiObject, passing that to the `<Form>`, manually creating a button and linking its onSubmit to the API? 

Edit: After an out-of-band sync, we decided to implement a feature for FormJ. This new feature should allow two new props that can be passed to the Form. They are injected into the result, allowing a user of formj, like suggest, to inject `<input type="submit">` buttons into the generated `<form>` tag.

PR #1270, opened 2026-04-14
Fix advanced generator test due to data change

baltpeter:
After https://github.com/datenanfragen/data/pull/3314, the suggested transport medium for `google` is now `webform`.

Let's instead use a record where we're sure that it will stay `suggested-transport-medium: email.


zner0L (approved):
Great, that should work!

PR #1278, opened 2026-06-05
Transparency update June 2026

baltpeter:
This should be squashed.

### DIFF
--- a/content/hr/contact.md
+++ b/content/hr/contact.md
@@ -19,7 +19,7 @@ Germany
 **Web:** [www.osobnipodaci.org/verein](https://www.osobnipodaci.org/verein)
 
 **Zakonski predstavnici:** Benjamin Altpeter i Lorenz Sieben  
-Datenanfragen.de e.&nbsp;V. je neprofitna organizacija i registrirana je udruga u okružnom sudu u Braunschweigu, pod matičnim brojem VR&nbsp;201732, i od porezne uprave Braunschweig-Wilhelmstraße potvrđena kao dobrotvorna organizacija.
+Datenanfragen.de e.&nbsp;V. je neprofitna organizacija i registrirana je udruga u okružnom sudu u Braunschweigu, pod matičnim brojem VR&nbsp;201732, i od porezne uprave Braunschweig-Wilhelmstraße potvrđena kao organizacija od javnog interesa.
 
 **Odgovornost za sadržaj prema članku 55. stavka 2. njemačkog državnog ugovora o medijima (MStV):**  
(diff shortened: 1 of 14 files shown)


content/hr/contact.md
```diff
@@ -19,7 +19,7 @@ Germany
 **Web:** [www.osobnipodaci.org/verein](https://www.osobnipodaci.org/verein)
 
 **Zakonski predstavnici:** Benjamin Altpeter i Lorenz Sieben  
-Datenanfragen.de e.&nbsp;V. je neprofitna organizacija i registrirana je udruga u okružnom sudu u Braunschweigu, pod matičnim brojem VR&nbsp;201732, i od porezne uprave Braunschweig-Wilhelmstraße potvrđena kao dobrotvorna organizacija.
+Datenanfragen.de e.&nbsp;V. je neprofitna organizacija i registrirana je udruga u okružnom sudu u Braunschweigu, pod matičnim brojem VR&nbsp;201732, i od porezne uprave Braunschweig-Wilhelmstraße potvrđena kao organizacija od javnog interesa.
```
zner0L, line 22: Why only this page, and only this language? Spanish has the same problems. Also, I am not sure of the translation, because this sounds like "public interest" more than public benefit.

zner0L (changes requested):
If you look into the translations of "public-benefit", this is great otherwise. Finally, we are back on track!

content/hr/contact.md
baltpeter: > Why only this page, and only this language?

Huh, I also changed this for other languages and pages where I caught it (e.g. https://github.com/datenanfragen/website/pull/1278/changes/d8016d4480e6cbde8b0ead161293b7f112735c23#diff-252f0e06943ea317665b8ae14c2b28eab1851774852203748e918009ae651f74R15)? :D

> Spanish has the same problems.

"organización benéfica" just sounded sufficiently public-benefit to my pattern matching, but you are right. That does appear to mean "charitable".

DeepL suggests "organización sin ánimo de lucro", whereas Google Translate says "organización de beneficio público". The latter seems more common per <https://www.linguee.com/english-spanish/search?query=public-benefit+organization>, so I'll go with that.

> Also, I am not sure of the translation, because this sounds like "public interest" more than public benefit.

Tjoa. "organizacija od javnog interesa" was suggested by DeepL, which indeed translates it back to "organization of public interest". Google instead suggests "organizacija za opće dobro", which DeepL translates to "public benefit organization", whereas Google translates it to "charity organization". *shrug*

baltpeter: Spanish is fixed.

zner0L (approved):
Thanks!

PR #1269, opened 2026-04-14
Replace Webpack, update to Node 24, Hugo 0.163.3, Cypress 15

baltpeter:
This should be rebase merged (once done).

PR #1279, opened 2026-06-16
Fix <table> structure in privacy controls

baltpeter:
cf. https://github.com/datenanfragen/website/pull/1269#pullrequestreview-4177020895

There was no reason for the \<div>. We do have some uses of .privacy-control but they work just as well on the \<tr>.


mal-tee:
Thank you!

This seems to change the position of the checkboxes, is that intended?

Currently:

grafik


New:
grafik

baltpeter:
Oops, that's embarrassing.

Issue #1271, opened 2026-04-15
Remove inline styles


mal-tee:
- [ ] refactor inline styles into proper css, etc.
- [ ] enable the `no-inline-styles` linter rule
- [ ] remove `unsafe-inline` from the CSP's style-src directive

PR #1283, opened 2026-07-19
Refactor inline style `float` and `margin-top` attributes into classes

mal-tee:
This should clean up the code base a bit. By removing inline-css we hopefully end up with more consistent code and layouts. :)  We should continue to do this with, e.g., the spacing attributes.

Additionally this works towards #1271 and a stronger CSP.


content/fr/verein/membership-fees.md
```diff
@@ -25,7 +25,7 @@ La classification en tant que mineur ou autres réductions ou exemptions sont to
 
 Les membres fondateurs peuvent choisir de ne pas payer la cotisation. Ils en sont explicitement exemptés par l'assemblée fondatrice.
 
-<div style="float: right;">
+<div class="float-right;">
```
baltpeter:
```suggestion
<div class="float-right">
```

assets/styles/framework.scss
```diff
@@ -111,6 +111,10 @@
     float: right;
 }
 
+.float-none {
```
baltpeter, line 114: Looks like we don't actually need that. Both uses have since been changed to use flexbox, where `float: none` has no effect.

baltpeter (changes requested):
It's too hot for me to continue reviewing but here's the review for the float stuff at least. :D

PR #1286, opened 2026-07-25
Remove references to deprecated Post product

mal-tee:


baltpeter:
Good catch! But seems like it was just renamed to [Einschreiben Wert](https://www.deutschepost.de/de/e/einschreiben/kuendigung-abo-themen/geld-verschicken.html), not actually discontinued.

mal-tee:
Oh! But I'd prefer to not link a specific product anyways :D It can't be outdated that way

baltpeter:
> Oh! But I'd prefer to not link a specific product anyways :D It can't be outdated that way

But the point of the FAQ item it to make this easy for people. Of course the information can go outdated, but so can everything else on the site. :D

mal-tee (commented):
ref to EINSCHREIBEN wert

baltpeter (approved):
Thank you!

